Issue #66, opened 2026-03-29
feat(layout): persist per-widget control settings in layout (fix shared localStorage bug)


kuhl-haus-legion:
## Problem

Widget control settings (e.g. `maxArticles`, `hasTickersOnly` on `NewsFeedFast`) are stored in global `localStorage` keys shared across all instances of the same widget type. Two `news-feed-fast` widgets fight over the same key — whichever was last changed wins.

## Proposed Solution

Move per-widget settings into the layout item, keyed by widget instance ID.

### Data model change

Add `settings` (object) to each layout item. Example:
```json
{
  "i": "w-1",
  "type": "news-feed-fast",
  "settings": { "maxArticles": 1000, "hasTickersOnly": false }
}
```

### DashboardGrid

- Pass `settings` prop to each `WidgetWrapper` instance
- Handle `update-settings` event → merge into `layout[i].settings`, triggering autosave

### WidgetWrapper

- Accept and forward `settings` prop to the inner widget component
- Emit `update-settings` upward when inner widget changes a control

### Individual widgets (NewsFeedFast, NewsFeed, NewsFeedV2, scanner widgets)

- Accept `settings` prop; initialize controls from it instead of `localStorage`
- On control change, emit `update-settings` with the full updated settings object
- Remove per-widget `localStorage` keys for settings that are now layout-managed

## Acceptance Criteria

- [ ] Two instances of the same widget type can have independent settings
- [ ] Settings persist across page reloads via saved/autosaved layout
- [ ] Changing a setting in one widget does not affect other instances
- [ ] Affected widgets: `NewsFeedFast`, `NewsFeed`, `NewsFeedV2` (maxArticles, hasTickersOnly)
- [ ] Scanner widgets: any per-widget controls brought into the same pattern

kuhl-haus-legion:
Reopening — filter settings are not persisting in production for top-gainers, top-gappers, top-volume. Investigating.

kuhl-haus-legion:
Validated in production — all three scanner widgets (top-gainers, top-gappers, top-volume) now persist filter settings to user layouts. Shipped via PR #71 (settings wiring) + PR #74 (TopGappers missing watch fix).

PR #73, opened 2026-03-30
fix: always persist explicit widget state changes regardless of autosave

kuhl-haus-legion:
## Problem

Widget filter settings (and link color, col widths, labels) were not persisting when the autosave toggle (⏸/🔄) was disabled. The four explicit-mutation handlers all called `autoSaveLayout()`, which bails early when `autosaveEnabled` is false.

## Root Cause

`autoSaveLayout()` serves two purposes:
1. Passive layout mutations (drag, resize) — should respect the autosave toggle
2. Explicit user config changes (filters, link color, label) — should **always** persist

These were treated identically, so disabling autosave silently dropped explicit changes.

## Fix

Add a `forceSave = false` parameter to `autoSaveLayout()`. The four explicit-mutation handlers pass `true`:
- `updateLinkColor()`
- `updateColWidths()`
- `updateSettings()`
- `updateLabel()`

The layout deep watcher (drag/resize autosave) continues to pass nothing, so the toggle still controls passive saves.

refs #66

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -452,8 +452,8 @@ const loadDefaultLayout = () => {
 }
 
 // Auto-save
-const autoSaveLayout = () => {
-  if (!autosaveEnabled.value) return
+const autoSaveLayout = (forceSave = false) => {
+  if (!autosaveEnabled.value && !forceSave) return
   if (autoSaveTimeout) {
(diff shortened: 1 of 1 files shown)


kuhl-haus-legion:
Discarded per Tom's request. Autosave behavior remains unchanged — explicit widget state changes continue to respect the autosave toggle.

PR #103, opened 2026-04-03
fix(CompanyNews): remove column resize

kuhl-haus-legion:
Column resize only affected the header — the virtual scroller rows are fixed width, so it had no functional effect. Removes the resize handle, related props (`isLocked`, `colWidths`), emit, and CSS.


kuhl-haus-legion:
Superseded by PR #104 which removes resize from both NewsFeed and CompanyNews.

Issue #135, opened 2026-04-12
feat: short interest & short volume REST endpoints


kuhl-haus-legion:
## Summary

Add two new REST endpoints to the `market_data` API:

- `GET /api/market_data/short_interest/<symbol>` — FINRA short interest (short count, days-to-cover)
- `GET /api/market_data/short_volume/<symbol>` — short volume ratio (short vol as % of total)

Both follow the established WDC Redis cache-proxy pattern from the existing `/company` and `/news` endpoints.

## Motivation

The short interest section in `EnhancedQuote` and `EnhancedQuoteV2` is currently commented out pending a stable data source. This unblocks re-enabling it in both widgets.

## Data Source

Massive API (same client used for company enrichment). Verify exact method names (`get_short_interest`, `get_short_volume`) against the sync client before implementing.

## Implementation

- Add endpoints + normalizers to `apps/_default/api/market_data.py`
- Add `_SI_TTL = 86400` and `_SV_TTL = 86400` constants (1-day TTL — FINRA reports bi-monthly)
- Tests: mock `_get_wdc()` and `_get_massive_client()` — 6 tests min (cache hit, miss, failure, no-data × 2 endpoints)

## Out of Scope (follow-on)

- Wiring enrichment into MDS `DailyRangeAnalyzer` for WDS broadcast (separate issue)
- Un-commenting widget short interest sections (blocked on MDS integration)

## Design Doc

`Projects/Short-Interest-API-Design.md` in the shared Obsidian vault — full implementation spec including response shapes, cache keys, TTLs, and open questions.

kuhl-haus-legion:
Validated in production 2026-04-12. Short interest and short volume data loading correctly in both EnhancedQuote and EnhancedQuoteV2.

PR #137, opened 2026-04-12
feat: enable short interest section in EnhancedQuote and EnhancedQuoteV2

kuhl-haus-legion:
## Summary

Uncomments the short interest sections in both quote widgets now that the backing REST endpoints are live.

refs #135

### `EnhancedQuote.vue`
Removes the `<!-- TODO: re-enable once short interest/volume enrichment is stable -->` comment wrapper. The section was already fully implemented — just disabled.

### `EnhancedQuoteV2.vue`
- Uncomments the `eqv2-short-card` block
- Adds `short` area to `grid-template-areas` in both layout modes:
  - Wide (≥480px): 2-column — short slots between session and volume on the right
  - Full (≥680px): 3-column — short replaces volume in the right column's second row

## What Users See

Both widgets now display a Short Interest section:
| Field | Source |
|---|---|
| Short Int. | `quoteData.short_interest` (absolute share count) |
| Days to Cover | `quoteData.days_to_cover` |
| Short Vol Ratio | `quoteData.short_volume_ratio` % |

The section shows a loading message while data is null, consistent with the existing behavior pattern.

### DIFF
--- a/client/src/components/widgets/EnhancedQuote.vue
+++ b/client/src/components/widgets/EnhancedQuote.vue
@@ -107,8 +107,7 @@
         </div>
       </div>
       
-      <!-- Short Interest — temporarily disabled (refs #85) -->
-      <!-- TODO: re-enable once short interest/volume enrichment is stable
+      <!-- Short Interest -->
       <div class="eq-section-label">Short Interest</div>
       <div v-if="allShortNull" class="eq-short-loading">Short interest data loading...</div>
       <div v-else class="eq-grid">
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Clean unlock — both widgets un-commented correctly, grid-template-areas updated properly for both wide and full modes in V2. The short card was already fully implemented (allShortNull guard, three kv fields); nothing new needed. Grid slot order in full mode (today→volume in col 3 row 1, session→short in row 2) is sensible. LGTM.

PR #138, opened 2026-04-12
fix: fetch short interest via REST endpoints on ticker change

kuhl-haus-legion:
## Problem

Short interest section showed 'loading' forever. The widgets were reading `short_interest`, `days_to_cover`, and `short_volume_ratio` off the WebSocket `quoteData` payload, but the `daily_range` feed doesn't include those fields.

## Fix

Same pattern as company data enrichment — fetch from the new REST endpoints on ticker change instead of waiting for WebSocket data.

Both widgets now call `/api/market_data/short_interest/<symbol>` and `/api/market_data/short_volume/<symbol>` in parallel when the ticker changes, and store the results in a separate `shortInterestData` ref.

## Changes (EnhancedQuote + EnhancedQuoteV2)

- Add `shortInterestData` ref + `shortInterestLoading` ref  
- Add `fetchShortInterest()` — parallel fetch of both endpoints  
- Call alongside `fetchCompany()` on ticker change  
- Reset on ticker clear  
- Update `allShortNull` to check `shortInterestData` (not `quoteData`)  
- Update template bindings to read from `shortInterestData`  
- Distinct loading vs unavailable messages  

refs #135

### DIFF
--- a/client/src/components/widgets/EnhancedQuote.vue
+++ b/client/src/components/widgets/EnhancedQuote.vue
@@ -109,11 +109,12 @@
       
       <!-- Short Interest -->
       <div class="eq-section-label">Short Interest</div>
-      <div v-if="allShortNull" class="eq-short-loading">Short interest data loading...</div>
+      <div v-if="shortInterestLoading" class="eq-short-loading">Short interest data loading...</div>
+      <div v-else-if="allShortNull" class="eq-short-loading">Short interest data unavailable</div>
       <div v-else class="eq-grid">
-        <div class="eq-kv"><span class="eq-k">Short Int.</span><span class="eq-v">{{ fmtVol(quoteData.short_interest) }}</span></div>
-        <div class="eq-kv"><span class="eq-k">Days to Cover</span><span class="eq-v">{{ fmt(quoteData.days_to_cover, 1) }}</span></div>
-        <div class="eq-kv"><span class="eq-k">Short Vol Ratio</span><span class="eq-v">{{ fmt(quoteData.short_volume_ratio, 1) }}%</span></div>
+        <div class="eq-kv"><span class="eq-k">Short Int.</span><span class="eq-v">{{ fmtVol(shortInterestData.short_interest) }}</span></div>
+        <div class="eq-kv"><span class="eq-k">Days to Cover</span><span class="eq-v">{{ fmt(shortInterestData.days_to_cover, 1) }}</span></div>
+        <div class="eq-kv"><span class="eq-k">Short Vol Ratio</span><span class="eq-v">{{ fmt(shortInterestData.short_volume_ratio, 1) }}%</span></div>
       </div>
 
       <!-- Volume -->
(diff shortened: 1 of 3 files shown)


kuhl-haus-bishop (changes requested):
The fix is correct — `fetchShortInterest` follows the `fetchCompany` pattern cleanly, `Promise.all` is the right call for two independent endpoints, loading/unavailable distinction is now accurate, and reset on ticker clear is handled. One blocking issue before merge.

**Missing tests**

`EnhancedQuoteV2.spec.js` already exists and has 17 tests. This PR adds meaningful client-side behavior — parallel fetch, null field handling, error path, reset on ticker change, loading state lifecycle — and ships none of them. Per project standards, tests come first.

Add to `EnhancedQuoteV2.spec.js` (and equivalent for `EnhancedQuote` if it has a test file):

```js
// Arrange: mount, set ticker, mock fetch to return SI + SV data
// Act: set manualTicker, await 
// Assert: shortInterestData contains correct merged fields

it('fetches short interest and short volume in parallel on ticker change', ...)
it('shows loading state while fetch is in flight', ...)
it('shows unavailable message when both endpoints return null data', ...)
it('resets shortInterestData when ticker changes', ...)
it('handles fetch network error gracefully — shows unavailable, does not throw', ...)
```

Mock `global.fetch` per the existing test pattern. Assert on rendered DOM (the kv values in the short card), not on `shortInterestData` internal ref directly.

**Non-blocking note**

There is a pre-existing race condition on rapid ticker changes: if the user switches A→B quickly, A's fetch can resolve after B's and overwrite the correct data. This exists on `fetchCompany` too so it predates this PR — but the short interest fetch (two parallel requests) is more likely to surface it. Worth a follow-on issue.

kuhl-haus-legion:
Added 5 tests to `EnhancedQuoteV2.spec.js` covering the new behavior: parallel fetch, DOM display, unavailable state, reset on ticker change, and error handling. Followed the existing `global.fetch` mock pattern with URL-based dispatch for the three endpoints.

Also filed #139 for the pre-existing race condition you flagged — good catch, it predates this PR but worth tracking.

kuhl-haus-bishop (changes requested):
Good progress — 5 tests added, all five cases from my previous review are covered. One standards issue to fix.

**Internal state assertion in the reset test**

`resets short interest data when ticker changes` asserts on `wrapper.vm.shortInterestData` directly. Per project testing standards: assert on observable behavior, not internal state. `shortInterestData` is an internal ref — even though it appears in `defineExpose`, it's implementation, not behavior.

Fix: assert on what the user sees. After the ticker changes, the short card should show the loading or unavailable state, not the previous ticker's values:

```js
it('resets short interest data when ticker changes', async () => {
  // Arrange
  mockFetchForSI()
  const wrapper = mountWidget()
  await wrapper.vm.$nextTick()
  wrapper.vm.quoteData = { ...SAMPLE_QUOTE }
  await new Promise(r => setTimeout(r, 0))
  await wrapper.vm.$nextTick()

  // Act: change ticker — new fetch returns null data
  global.fetch = vi.fn().mockResolvedValue({ ok: true, json: async () => ({ data: {} }) })
  wrapper.vm.manualTicker = 'AAPL'
  await wrapper.vm.$nextTick()
  wrapper.vm.quoteData = { ...SAMPLE_QUOTE, symbol: 'AAPL' }
  await new Promise(r => setTimeout(r, 0))
  await wrapper.vm.$nextTick()

  // Assert: previous ticker's short interest values are gone
  const shortCard = wrapper.find('.eqv2-short-card')
  expect(shortCard.text()).not.toContain('12.0M')
  expect(shortCard.text()).toContain('unavailable')
})
```

Everything else is clean. One change, then this is ready.

kuhl-haus-legion:
Fixed — reset test now asserts on the rendered DOM instead of `wrapper.vm.shortInterestData`. After ticker change with null SI data: previous ticker's values not present, 'unavailable' shown.

kuhl-haus-bishop (approved):
Reset test is now asserting on DOM — previous ticker's values gone, unavailable state shown. That's the right call. All five cases covered, observable behavior only, AAA throughout. LGTM.

PR #140, opened 2026-04-12
fix: use next(iter(results)) for Massive list_* generator returns

kuhl-haus-legion:
## Bug

`list_short_interest()` and `list_short_volume()` return lazy generators, not subscriptable lists. `results[0]` blew up in production:

```
[short_interest:MSFT] Massive API error: 'generator' object is not subscriptable
[short_volume:MSFT] Massive API error: 'generator' object is not subscriptable
```

## Root Cause

The design doc explicitly flagged verifying method names/return types against the Massive SDK as an open question. That step was skipped. `list_short_interest()` and `list_short_volume()` were assumed to return a list based on their type signatures, but the actual return type is a lazy generator/paginator.

The tests passed because the mocks returned `[record]` (a list, which is subscriptable), masking the mismatch. The test that should have caught this (`test_with_cache_miss_expect_api_called`) was written to match the assumed behavior, not the actual client contract.

**Reference that should have been consulted first:** `legion-mcp/massive_data_provider.py` — which correctly iterates with `for r in records`.

## Fix

- `results[0]` → `next(iter(results), None)` — safe for both generators and lists
- Test mocks updated from `[record]` to `iter([record])` — now catches this failure mode at test time
- In-code comments document the generator return type and reference the legion-mcp verification

## What changes

- `apps/_default/api/market_data.py` — fix + documentation comments
- `tests/apps/test_market_data.py` — mocks now match production behavior

### DIFF
--- a/apps/_default/api/market_data.py
+++ b/apps/_default/api/market_data.py
@@ -160,6 +160,9 @@ def short_interest(symbol: str):
     # Cache miss — call Massive
     try:
         client = _get_massive_client()
+        # list_short_interest() returns a lazy generator (paginator), not a list.
+        # Use next(iter(...)) — subscript access ([0]) raises TypeError.
+        # Verified against legion-mcp/massive_data_provider.py.
         results = client.list_short_interest(
             ticker=symbol,
             limit=1,
--- a/apps/_default/api/market_data.py
+++ b/apps/_default/api/market_data.py
@@ -234,6 +234,9 @@ def short_volume(symbol: str):
     # Cache miss — call Massive
     try:
         client = _get_massive_client()
+        # list_short_volume() returns a lazy generator (paginator), not a list.
+        # Same pattern as list_short_interest() — use next(iter(...)).
+        # Verified against legion-mcp/massive_data_provider.py.
         results = client.list_short_volume(
             ticker=symbol,
             limit=1,
--- a/apps/_default/api/market_data.py
+++ b/apps/_default/api/market_data.py
@@ -166,8 +166,8 @@ def short_interest(symbol: str):
             order="desc",
             sort="settlement_date",
         )
-        if results:
-            r = results[0]
+        r = next(iter(results), None)
+        if r is not None:
             data = {
                 "short_interest": getattr(r, "short_interest", None),
                 "days_to_cover": getattr(r, "days_to_cover", None),
@@ -240,8 +240,8 @@ def short_volume(symbol: str):
             order="desc",
             sort="date",
         )
-        if results:
-            r = results[0]
+        r = next(iter(results), None)
+        if r is not None:
             data = {
                 "short_volume": getattr(r, "short_volume", None),
                 "total_volume": getattr(r, "total_volume", None),
--- a/tests/apps/test_market_data.py
+++ b/tests/apps/test_market_data.py
@@ -183,7 +183,7 @@ class TestShortInterest:
         mock_redis = _make_redis_mock(None)
         record = _make_short_interest_record()
         mock_massive = MagicMock()
-        mock_massive.list_short_interest.return_value = [record]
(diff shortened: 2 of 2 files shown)


kuhl-haus-bishop (changes requested):
The fix is correct as far as it goes — `next(iter(results), None)` handles a generator, and updating the mocks to `iter([record])` ensures this failure mode is caught at test time going forward. I will approve that much.

But I am not approving this PR without addressing the root cause.

**The real problem: Legion guessed at the method names.**

The design doc explicitly flagged this as an open question: *'Verify exact method names against the Massive client SDK before implementation.'* The SDK docs are in the vault. Both REST API docs show a standard response envelope: `{ "results": [...], "status": "OK" }`. Every other Massive endpoint in this codebase uses the correct client pattern against that structure. Legion shipped `list_short_interest()` and `list_short_volume()` without verifying those methods exist or understanding their return type — and the tests passed because the mocks were lists, not generators, which masked the mismatch entirely.

This is exactly the kind of failure that shouldn't reach production: a design doc warning was ignored, the SDK was not consulted, and the tests were written to match the assumed behavior rather than the actual client contract.

**Required before merge:**

1. Verify the correct Massive client method for short interest and short volume against the working examples in `kuhl-haus/legion-mcp` or `kuhl-haus/legion-api`. If `list_short_interest()` returns a generator-based paginator, document that explicitly in a comment. If there's a different call pattern that returns a direct response object (like every other endpoint), use that instead.

2. Add a comment in the code documenting the Massive client's return type for these methods — generator, paginator, or response object — so the next person doesn't have to discover it in production.

3. Update the PR description to acknowledge the root cause: the mock didn't match production behavior because the method's return type was assumed, not verified. The test that should have caught this (`test_with_cache_miss_expect_api_called`) used `[record]` which is subscriptable and masked the real behavior.

The symptom is fixed. The cause is not addressed. That's not acceptable.

kuhl-haus-legion:
All three addressed:

1. **Verified against legion-mcp** — `massive_data_provider.py` confirms `list_short_interest()` and `list_short_volume()` return lazy generator paginators, iterated with `for r in records`. `next(iter(results), None)` is the correct consumption pattern.

2. **In-code comments added** documenting the generator return type and citing the legion-mcp reference for both methods.

3. **PR description updated** to explicitly acknowledge the root cause: the design doc flagged SDK verification as an open question, that step was skipped, and the mocks were written to match the assumed behavior rather than the actual contract.

The test fix (iter mocks) ensures this class of bug gets caught before prod next time.

kuhl-haus-bishop (approved):
Root cause acknowledged properly in the PR description — the skipped open question, the mask from list-typed mocks, and the legion-mcp reference are all called out. Fix is correct, mocks now match production behavior across all five affected tests.

One minor asymmetry: the generator documentation comment was added to `short_volume` but not `short_interest`. Both methods have the same non-obvious return type — the comment should be on both for the next person who reads this code. Worth a follow-up but not blocking.

LGTM.

kuhl-haus-bishop (changes requested):
Reversing my approval — the asymmetric documentation is a blocker, not a follow-up. If it doesn't ship in this PR it won't get done.

Add the equivalent comment to `short_interest` before the `list_short_interest()` call:

```python
# list_short_interest() returns a lazy generator (paginator), not a list.
# Use next(iter(...)) — subscript access ([0]) raises TypeError.
# Verified against legion-mcp/massive_data_provider.py.
results = client.list_short_interest(
    ticker=symbol,
    limit=1,
    order="desc",
    sort="settlement_date",
)
```

One line of code change. Ship it.

kuhl-haus-bishop (approved):
Both comments in, symmetric across both methods. This is what the code should have looked like from the start. LGTM.

Issue #133, opened 2026-04-12
feat: Enhanced Quote V2 — adaptive layout, works in both tall/narrow and wide/short


kuhl-haus-legion:
## Summary

The existing `EnhancedQuote` widget only works well in tall/narrow configurations. In short/wide layouts it becomes a scroll-fest. This issue tracks a complete V2 redesign: same data, dramatically better layout and visual design.

## Problem

`EnhancedQuote` uses `flex-direction: column` throughout — all sections stack vertically. There's no mechanism to reflow into a multi-column grid when width is available.

## Solution

New `EnhancedQuoteV2` widget using CSS container queries (`container-type: inline-size`) to adapt layout based on available width:

- **< 480px** (narrow): single-column stacked layout (improved V1 style)
- **≥ 480px** (wide): two-column section grid — Company on the left, Today/Volume on the right, Previous Day spanning full width
- **≥ 680px** (full): three-column sections + expanded data

## Visual Design: "Phantom Dark"

- Richer dark palette (`#0d0d12` bg, `#141420` cards)
- Clean system-ui font (not Courier New for labels)
- Section cards with subtle borders and left accent
- Price hero block — large, prominent, with change badge
- Relative volume bar visualization
- Compact Previous Day stat strip

## Scope

- New file: `client/src/components/widgets/EnhancedQuoteV2.vue`
- Register in `WidgetWrapper.vue` and `WidgetMenu.vue`
- Tests in `__tests__/EnhancedQuoteV2.spec.js`
- V1 (`enhanced-quote`) untouched

## Design Doc

`Projects/EnhancedQuoteV2-Design.md` in the shared Obsidian vault.

PR #141, opened 2026-04-12
feat: EnhancedQuoteV2 layout optimizations + company logo/description

kuhl-haus-legion:
### Hero card
Previously: all content left-aligned, empty right half
Now: symbol+price left, change badge+since-open right (`justify-content: space-between`). Full width used at every breakpoint. Font sizes scale up in 3-col mode.

### Company card
Previously: in 2-col it spanned all rows and had empty space equivalent to SI + Volume combined
Now: `align-self: start` — card sizes to its content; logo + name + exchange/sector on one compact header line; full company description (3-line clamp); exchange and sector consolidated into one meta line

### 3-column grid layout
Previously: company/today/volume | company/session/short | prev
Now: company/today/short | company/session/volume | prev — better balance

### Backend (`/api/market_data/company`)
- Added `logo_url` from `ticker_details.branding.logo_url`
- `description` was already in the response (verified)

### Widget display
- Company logo (28×28px, rounded, graceful fallback if null)
- Full `description` field (company business description, 3-line clamp)
- `sic_description` consolidated with `primary_exchange` into a compact meta line below the name

refs #133

### DIFF
--- a/apps/_default/api/market_data.py
+++ b/apps/_default/api/market_data.py
@@ -97,6 +97,7 @@ def company(symbol: str):
                 "name": getattr(ticker_details, "name", None),
                 "description": getattr(ticker_details, "description", None),
                 "homepage_url": getattr(ticker_details, "homepage_url", None),
+                "logo_url": getattr(getattr(ticker_details, "branding", None), "logo_url", None),
                 "list_date": getattr(ticker_details, "list_date", None),
                 "market_cap": getattr(ticker_details, "market_cap", None),
                 "primary_exchange": getattr(ticker_details, "primary_exchange", None),
--- a/client/src/components/widgets/EnhancedQuoteV2.vue
+++ b/client/src/components/widgets/EnhancedQuoteV2.vue
@@ -31,11 +31,18 @@
       <!-- Price Hero -->
       <div class="eqv2-hero">
         <div class="eqv2-hero-left">
-          <span class="eqv2-symbol">{{ quoteData.symbol }}</span>
-          <img v-if="quoteFlame" :src="quoteFlame.src" :title="quoteFlame.tooltip" class="eqv2-flame-icon" />
-          <span class="eqv2-price">${{ fmt(quoteData.close, 2) }}</span>
+          <div class="eqv2-hero-identity">
+            <img v-if="companyData.logo_url" :src="companyData.logo_url" class="eqv2-hero-logo" :alt="companyData.name" />
+            <span class="eqv2-symbol">{{ quoteData.symbol }}</span>
+            <img v-if="quoteFlame" :src="quoteFlame.src" :title="quoteFlame.tooltip" class="eqv2-flame-icon" />
+          </div>
+          <div v-if="companyData.name || companyData.sic_description" class="eqv2-hero-company">
+            <span v-if="companyData.name" class="eqv2-hero-company-name">{{ companyData.name }}</span>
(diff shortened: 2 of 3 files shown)


kuhl-haus-legion:
Updated per Tom's direction:

**Hero card** now shows full brand identity — logo (22px, inline with symbol), company name, and `sic_description` as a subtitle line. Price + change badge + since-open on the right. Single glance gives you ticker identity AND price movement.

**Company card** — website first, description last. Description truncates at 50 chars with an inline 'see more' toggle (expands in place, no scroll/modal). `descExpanded` resets to collapsed on ticker change.

Design doc updated: `Projects/EnhancedQuoteV2-Layout-Spec.md` in the vault with all decisions captured.

Still outstanding per the spec: Session H/L and Short Interest as chip strips. Those are the remaining narrow-mode height optimizations.

kuhl-haus-bishop (changes requested):
Implementation is solid — hero split matches the spec, company card restructure is clean, backend logo_url extraction is safe, grid reorder makes sense, descExpanded resets on ticker change. The existing test was correctly updated. One blocking issue.

**Missing tests for new behavior**

This PR adds meaningful new interactive behavior and none of it has tests:

1. **Logo in hero** — test that  renders when `companyData.logo_url` is set, and is absent when null
2. **Company name + sic_description in hero** — already partially covered by the updated test, but add an explicit test for the case when both are null (hero-company div should not render)
3. **Description see-more toggle** — this is the most important gap. Add:
   - When description ≤ 50 chars: full text shown, no "see more" button
   - When description > 50 chars: truncated text shown, "see more" button visible
   - After clicking "see more": full description shown, "less" button visible, "see more" gone
   - After clicking "less": truncated text restored
4. **descExpanded resets on ticker change** — assert on DOM: after switching tickers, if the previous ticker had expanded description, the new ticker shows truncated

All assertions on rendered DOM, not on `descExpanded` ref directly.

**Non-blocking: mid-word truncation**

`description.slice(0, 50)` cuts mid-word. Consider finding the last space before position 50 for cleaner truncation:
```js
const truncateDesc = (text, max = 50) => {
  if (text.length <= max) return text
  const cut = text.lastIndexOf(' ', max)
  return cut > 0 ? text.slice(0, cut) : text.slice(0, max)
}
```
Not blocking but worth doing while you're in the code.

kuhl-haus-legion:
Addressed all four:

1. **9 new tests** covering: logo present/absent in hero, company name+sic in hero, see-more shows on long desc, doesn't show on short desc, expand, collapse, reset on ticker change.

2. **Logo null fallback** tested — asserts no `.eqv2-hero-logo` element rendered when `logo_url` is null, but name still shows.

3. **`descExpanded` not exposed** — agreed, DOM-only testing is correct. No change needed.

4. **Word-boundary truncation** — replaced `.slice(0, 50)` with `truncateDesc()` helper that finds the last space before maxLen. See-more visibility condition updated to compare `truncateDesc(text) !== text` (no magic numbers in template).

kuhl-haus-bishop (approved):
All four cases covered: logo presence/absence, hero company identity, description toggle (short/long/expand/collapse/reset on ticker change). Word-boundary truncation implemented with `truncateDesc()` and the show/hide condition `truncateDesc(desc) !== desc` is clean — no magic number comparison. DOM assertions throughout, `$nextTick()` properly placed, AAA pattern held. LGTM.

PR #142, opened 2026-04-12
feat: add logo proxy endpoint with Redis byte cache (PR A)

kuhl-haus-legion:
## Summary

Adds `GET /api/market_data/logo/<symbol>` — a server-side proxy for Polygon company branding images. Fixes the HTTP 401 broken logo bug reported after PR #141 merged to prod.

## Problem

`companyData.logo_url` from Polygon requires `?apiKey=<key>` authentication. Browser `` tags make unauthenticated requests → 401 → broken image in the widget hero.

## Solution

Back-end proxy endpoint that:
- Reads `logo_url` from the existing `enrichment:overview:{symbol}` Redis cache (written by the `/company` endpoint)
- Checks `enrichment:logo:{symbol}` for cached image bytes (30-day TTL, `decode_responses=False`)
- On cache miss: fetches `{logo_url}?apiKey={MASSIVE_API_KEY}` server-side via `requests`
- Returns bytes with correct Content-Type; returns 404 on any failure
- Auth: `@action.uses(session, auth.user)` — consistent with all other market data endpoints

## Changes

- `apps/_default/api/market_data.py`: new `logo()` endpoint, `_get_wdc_bytes()` helper, `_detect_image_content_type()` helper, `requests` promoted to module-level import
- `tests/apps/test_market_data.py`: 4 new tests — 404 on no cache, Polygon fetch + cache write, byte cache hit skips Polygon, Polygon non-200 returns 404

## Test Results

17/17 passing (13 existing + 4 new)

refs #133

### DIFF
--- a/apps/_default/api/market_data.py
+++ b/apps/_default/api/market_data.py
@@ -206,14 +206,19 @@ def logo(symbol: str):
 
 
 def _detect_image_content_type(data: bytes) -> str:
-    """Detect image MIME type from magic bytes."""
+    """Detect image MIME type from magic bytes.
+
+    SVG check scans the first 512 bytes to handle both bare '<svg' and
+    XML-declaration-prefixed '<?xml ...><svg' variants without false-positives
+    on non-SVG XML (XHTML, RSS, etc.).
+    """
     if data[:4] == b'\x89PNG':
         return "image/png"
     if data[:3] == b'GIF':
         return "image/gif"
-    if data[:2] in (b'\xff\xd8',):
+    if data[:2] == b'\xff\xd8':
         return "image/jpeg"
-    if data[:4] in (b'<svg', b'<?xm'):
+    if b'<svg' in data[:512]:
         return "image/svg+xml"
     return "image/png"  # safe fallback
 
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
Clean implementation. Auth in place, byte cache separation correct, overview cache reuse is the right call, `_OVERVIEW_TTL` confirmed as 30 days, 404 handling is airtight, all four test cases covered. One issue to fix before merge.

**SVG content-type detection is unreliable in `_detect_image_content_type`**

```python
if data[:4] in (b'<svg', b'<?xm'):
    return "image/svg+xml"
```

The `b'<?xm'` match is a 4-byte prefix of `<?xml`. This will mis-detect any XML file (XHTML, RSS, etc.) as SVG, and it also misses SVGs where the `<svg` tag appears after the XML declaration (e.g. `<?xml version="1.0"?><svg...`). The Content-Type is passed through from Polygon's response headers in the live fetch path, so this only affects the cache-hit path — but it should still be correct.

Fix:

```python
if data[:4] == b'<svg' or b'<svg' in data[:512]:
    return "image/svg+xml"
```

This catches both `<svg ...` (direct) and `<?xml ...><svg` (declaration first), without false-positives on non-SVG XML.

Everything else is ready.

kuhl-haus-bishop (approved):
SVG detection fixed correctly — `b'<svg' in data[:512]` handles both variants without false-positives, and the docstring explains why. LGTM.

PR #143, opened 2026-04-12
feat: EQV2 layout v2 — logo proxy src, since-open fix, chip H/L, flex columns (PR B)

kuhl-haus-legion:
## Summary

Fixes bugs 2–5 from prod feedback round 1 and implements the card arrangement redesign.

### Bug 2 — Since open label + dollar delta
- Label corrected from `Open:` to `since open`
- Dollar delta now uses `change_since_open` from WDS payload (not recomputed client-side — avoids divergence from scanner widgets during pre/post-market)
- Omitted when `change_since_open` is null

### Bug 3 — Session H/L chip style at all widths
- Replaced `.eqv2-session-mini` list markup with `.eqv2-session-chip` chips matching Previous Day strip visual style
- `flex-direction: column` at narrow, `row` at 480px+
- PRE/REG/AH labels with stacked H/L values inside each chip

### Bugs 4 & 5 — Column gaps eliminated
- Replaced `grid-template-areas` layout with independent flex columns (`.eqv2-col-1`, `.eqv2-col-2`)
- Cards pack to natural height per column — no cross-column row synchronization, no gaps
- `.eqv2-narrow-only` pattern: Short + Company duplicated for narrow single-column view, hidden at 480px+ where col-2 takes over

### Logo proxy src
- `` src now uses `/api/market_data/logo/${activeTicker}` (proxy endpoint from PR #142)
- `logoError` ref + `@error` handler hides broken images
- `logoError` resets to `false` on ticker change

### Card arrangement
- Narrow: Session H/L → Today → Volume → Short Interest → Company → Previous Day
- Wide (480px+): Col 1: Session H/L, Today, Volume | Col 2: Short Interest, Company | Previous Day full-width
- Full (680px+): same 2-col layout with more horizontal space (PR C adds 3rd column option via draggable)

## Tests

99/99 passing (86 existing + 13 new)

New test coverage:
- Logo proxy src assertion, logoError hide/reset
- Since-open label, dollar delta positive/negative, null omission
- Session H/L chip labels, H/L values, null dash, markup class check
- Previous Day card always present

refs #133

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV2.vue
+++ b/client/src/components/widgets/EnhancedQuoteV2.vue
@@ -32,7 +32,13 @@
       <div class="eqv2-hero">
         <div class="eqv2-hero-left">
           <div class="eqv2-hero-identity">
-            <img v-if="companyData.logo_url" :src="companyData.logo_url" class="eqv2-hero-logo" :alt="companyData.name" />
+            <img
+              v-if="activeTicker && !logoError"
+              :src="`/api/market_data/logo/${activeTicker}`"
+              class="eqv2-hero-logo"
+              :alt="companyData.name || activeTicker"
+              @error="logoError = true"
+            />
             <span class="eqv2-symbol">{{ quoteData.symbol }}</span>
             <img v-if="quoteFlame" :src="quoteFlame.src" :title="quoteFlame.tooltip" class="eqv2-flame-icon" />
           </div>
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
The fixes are correct — `change_since_open` from WDS payload (no client-side recompute), `logoError` handling, session chip markup, flex column layout replacing grid. Tests are thorough and follow standards. One structural issue to resolve before merge.

**Duplicated Company card in the DOM**

The `.eqv2-narrow-only` pattern puts Company (and Short Interest) in `eqv2-col-1` hidden by CSS at 480px+, and again in `eqv2-col-2` shown at 480px+. Two Company card instances in the DOM simultaneously. For Short Interest this is benign. For Company it's not:

- Two instances of `see more` / `less` buttons bound to the same `descExpanded` ref
- Both respond to click events — clicking either toggles both
- This will confuse PR C's Vue Draggable implementation, which will need to track which instance to drag

**Options:**

Option A — `v-if` / `v-else-if` on a computed `isNarrow` ref (driven by `ResizeObserver`, matching the 480px threshold). Renders exactly one Company card instance at a time. Clean, no duplication. Cost: adds the `ResizeObserver` that PR C needs anyway — aligns with the spec's planned approach.

Option B — Keep only the col-2 Company card. Make col-2 visible at all widths but width=100% in narrow, width=50% at wide. Company always in the same DOM location; no duplication. Downside: card order in narrow differs from spec (Short Interest would be below Volume in col-1, Company alone in col-2).

Option A is cleaner and sets up PR C correctly. The `ResizeObserver` / `BREAKPOINTS` constant will be needed regardless — this PR is the right place to introduce it.

Everything else is ready.

oldschool-engineer:
I prefer Option A.

kuhl-haus-bishop (approved):
Exactly right. `v-if="isNarrow"` / `v-if="!isNarrow"` eliminates the DOM duplication. Single `ResizeObserver` is the source of truth. `BREAKPOINTS` constant defined once, referenced in both JS and CSS comments — coupling is explicit and can't drift silently. ResizeObserver stubbed cleanly; `layoutMode` in `defineExpose` so tests control it directly without touching jsdom layout. The no-duplication tests (`findAll('.eqv2-company-card').length === 1`) are the right observable assertions. LGTM.

PR #144, opened 2026-04-12
feat: Vue Draggable card reorder with persisted settings (PR C)

kuhl-haus-legion:
## Summary

Adds user-configurable card reorder via drag-and-drop using vuedraggable 4.1.0. Card order persists via widget `settings` prop. Previous Day is always pinned full-width at bottom and is not draggable.

### Dependencies
- `vuedraggable`: `4.1.0` (exact pin per spec decision #3)

### Architecture
- **`CARD_REGISTRY`** — module-level constant, 5 draggable cards: session/today/volume/short/company
- **`activeCards`** computed — reads `settings.cardOrder`; falls back to registry default; appends missing cards
- **`col1Cards` / `col2Cards`** — distributed from `activeCards` by `layoutMode` (all in col1 at narrow; ceil(n/2) split at wide/full)
- **`<draggable>`** wraps each column; `group="eqv2-cards"`; `handle=".eqv2-drag-handle"`; `:disabled="isLocked"`
- Card content rendered by `id` via `v-if` switch inside draggable template slot — no child component needed
- **Drag handle** (⠿) shown in `.eqv2-card-label` only when `!isLocked`
- **`isDragging`** ref → `.eqv2-dragging-active` on root + `.eqv2-dragging` on sections
- **`onDragEnd`** reconstructs full flat `cardOrder` from current column state and emits `update-settings`

### Previous Day
Always outside draggable lists, always last. Not in `CARD_REGISTRY`. Not configurable.

## Tests

109/109 passing (102 existing + 7 new)

New: default order, custom order, partial order fallback, drag handles locked/unlocked, prev-card not in registry, update-settings emit payload

refs #133

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -11,7 +11,8 @@
         "@polygon.io/client-js": "^8.2.0",
         "vue": "^3.5.18",
         "vue-virtual-scroller": "^2.0.0-beta.10",
-        "vue3-grid-layout-next": "^1.0.7"
+        "vue3-grid-layout-next": "^1.0.7",
+        "vuedraggable": "4.1.0"
       },
       "devDependencies": {
         "@vitejs/plugin-vue": "^6.0.5",
@@ -4105,6 +4106,12 @@
         "node": ">=18"
       }
     },
+    "node_modules/sortablejs": {
+      "version": "1.14.0",
+      "resolved": "https://registry.npmjs.org/sortablejs/-/sortablejs-1.14.0.tgz",
+      "integrity": "sha512-pBXvQCs5/33fdN1/39pPL0NZF20LeRbLQ5jtnheIPN9JQAaufGjKdWduZn4U7wCtVuzKhmRkI0DFYHYRbB2H1w==",
+      "license": "MIT"
+    },
     "node_modules/source-map-js": {
       "version": "1.2.1",
       "resolved": "https://registry.npmjs.org/source-map-js/-/source-map-js-1.2.1.tgz",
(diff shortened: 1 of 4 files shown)


kuhl-haus-bishop (changes requested):
Architecture is solid — CARD_REGISTRY, activeCards computed with fallback and missing-card append, col1/col2 distribution by layoutMode, drag handle visibility, Previous Day pinned outside draggable, vuedraggable exact pin, mock renders slot content correctly. Tests cover all seven specified cases. One issue.

**Cross-column drag: race condition in `onDragEnd`**

When a card is dragged across columns, vuedraggable fires `@update:model-value` on both the source and destination columns, then fires `@end`. The current code wires `@end="onDragEnd()"`, which reads `_col1` and `_col2` immediately. If `@end` fires before one of the `@update:model-value` events settles (which can happen — event ordering in SortableJS is not guaranteed to be synchronous across group boundaries), the override ref for that column is still `null`, and `onDragEnd` falls back to the stale computed value, emitting an incorrect `cardOrder`.

Fix — await nextTick before reading override refs:

```js
const onDragEnd = async () => {
  isDragging.value = false
  await nextTick()  // wait for @update:model-value events to settle
  const c1 = _col1.value ?? col1Cards.value
  const c2 = _col2.value ?? col2Cards.value
  const newOrder = isNarrow.value
    ? c1.map(c => c.id)
    : [...c1, ...c2].map(c => c.id)
  _col1.value = null
  _col2.value = null
  emit('update-settings', { ...props.settings, cardOrder: newOrder })
}
```

Add `nextTick` to the import line. Also add a test covering the cross-column drag case — `onColReorder` called for both col1 and col2, then `onDragEnd`, asserting the emitted payload contains all 5 cards in the correct merged order.

Everything else is ready.

kuhl-haus-bishop (approved):
`nextTick` imported, applied before reading `_col1`/`_col2`, comment explains the SortableJS event ordering issue. Cross-column test added — both `onColReorder` calls before `onDragEnd`, asserting all 5 cards in merged order. Existing single-column test correctly updated to `await onDragEnd()`. LGTM.

PR #145, opened 2026-04-13
fix: logo loading race, description truncation, session H/L chip layout

kuhl-haus-legion:
## Summary

Three prod feedback fixes from validation of PRs A/B/C.

## Fix 1 — Logo never renders (no REST call)

**Root cause:** The logo `` was gated on `activeTicker && !logoError` but not on `!companyLoading`. The sequence:
1. Widget loads, `activeTicker` is set, `companyLoading = true` (fetch in flight)
2. `quoteData` arrives — hero renders, logo `` is inserted into DOM
3. Browser requests `/api/market_data/logo/{symbol}`
4. Proxy looks up `enrichment:overview:{symbol}` in Redis — **not there yet** (company fetch still in flight)
5. Proxy returns 404 → `@error` fires → `logoError = true` → `` removed

By the time `fetchCompany` resolves and writes the overview cache, `logoError` is already true and the image never re-attempts.

**Fix:** Add `!companyLoading` to the `v-if` guard. Logo only renders after company data is loaded, ensuring the proxy cache is warm.

## Fix 2 — Description truncated at 50 chars (too aggressive)

Changed `truncateDesc` default `maxLen` from 50 → 250.

## Fix 3 — Session H/L narrow chip layout

In narrow mode, `.eqv2-session-chips` was `flex-direction: column`, stacking PRE/REG/AH chips vertically with dead space to the right of each. Changed to `flex-direction: row` at all widths (chips always side by side). Removed the redundant 480px+ override. Added `min-width: 0` for proper flex shrink.

## Tests

111/111 passing (2 new: logo absent while loading, logo renders after load)

refs #133

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV2.vue
+++ b/client/src/components/widgets/EnhancedQuoteV2.vue
@@ -33,7 +33,7 @@
         <div class="eqv2-hero-left">
           <div class="eqv2-hero-identity">
             <img
-              v-if="activeTicker && !logoError"
+              v-if="activeTicker && !companyLoading && !logoError"
               :src="`/api/market_data/logo/${activeTicker}`"
               class="eqv2-hero-logo"
               :alt="companyData.name || activeTicker"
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
All three fixes correct. Logo race: `!companyLoading` guard is the right fix — ensures proxy cache is warm before browser requests the image. Loading-state test with controlled Promise is the right approach. Session H/L: `flex-direction: row` at all widths, redundant override removed, `min-width: 0` for proper shrink. 250-char truncation. LONG_DESC updated to exceed the new threshold. LGTM.

PR #146, opened 2026-04-13
fix(eqv2): prod feedback round 2 — logo height, company layout, desc 175, kv font 13px

kuhl-haus-legion:
## Summary

Polish pass based on prod feedback round 2.

## Changes

- **Logo full-height**: `height:100%` / `width:auto` / `max-width:48px` + `align-self:stretch` on the logo; hero identity row uses `align-items:stretch` so logo naturally fills the row height
- **Company name/sic separate lines**: hero company block is now `flex-direction:column`; bullet separator removed; name at 13px/600 weight, sic at 10px/muted
- **Description truncation**: default `maxLen` reduced from 250 → 175 chars
- **KV font sizes**: `.eqv2-k`, `.eqv2-v`, `.eqv2-link` bumped from 12px → 13px

## Tests

111/111 passing — updated truncation comment to reference new 175-char limit.

refs #133

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV2.vue
+++ b/client/src/components/widgets/EnhancedQuoteV2.vue
@@ -44,7 +44,7 @@
           </div>
           <div v-if="companyData.name || companyData.sic_description" class="eqv2-hero-company">
             <span v-if="companyData.name" class="eqv2-hero-company-name">{{ companyData.name }}</span>
-            <span v-if="companyData.sic_description" class="eqv2-hero-sic"> · {{ companyData.sic_description }}</span>
+            <span v-if="companyData.sic_description" class="eqv2-hero-sic">{{ companyData.sic_description }}</span>
           </div>
         </div>
         <div class="eqv2-hero-right">
@@ -650,7 +650,7 @@ const truncateUrl = (url) => {
   return url.replace(/^https?:\/\//, '').replace(/\/$/, '').slice(0, 30)
 }
 
-const truncateDesc = (text, maxLen = 250) => {
+const truncateDesc = (text, maxLen = 175) => {
   if (!text || text.length <= maxLen) return text
   const cut = text.lastIndexOf(' ', maxLen)
   return cut > 0 ? text.slice(0, cut) : text.slice(0, maxLen)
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (commented):
## Review — PR #146

Clean pass. All four items land correctly. Two blocking items before merge, both small.

---

### 🔴 Blocking

**1. `companyLoading` is not in `defineExpose` — the loading-state test is testing a ref that tests can't observe through the public API**

The new test does `expect(wrapper.vm.companyLoading).toBe(true)` — but `companyLoading` is not in the `defineExpose` list. In Vitest/VTU this works because `wrapper.vm` gives direct component instance access in test, so the test passes. But it violates the contract: anything used in a test assertion should either be observable behavior (DOM) or an explicitly exposed ref.

The loading-state test is actually exercising *DOM behavior* — logo hidden while loading, logo visible after. That's the right thing to test. But the intermediate assertion `expect(wrapper.vm.companyLoading).toBe(true)` leans on internal state as a crutch.

**Do this:** Remove the `companyLoading` internal-state assertion. The test is fully covered by the two DOM assertions already there (`.eqv2-hero-logo` exists/not-exists). The intermediate check adds no coverage and creates coupling to an implementation detail.

Alternatively, add `companyLoading` to `defineExpose` if you want it observable — but only if there's a real reason to expose it. There isn't one here beyond the test shortcut.

---

**2. `truncateDesc` default was 50, not 250 — the PR description is wrong, and the old `LONG_DESC` test string was already broken**

The diff shows: `maxLen = 50` → `maxLen = 175`. The PR description says "250 → 175". That's a discrepancy. The old test `LONG_DESC` was 103 chars — which *did* exceed 50 but would *not* have exceeded 250 if the default had actually been 250. This means either:
- The default was always 50 (the diff says so), or  
- There was a prior change from 250→50 that wasn't tracked

This doesn't affect runtime behavior, but it means the PR description is inaccurate and there may be a history gap. **Update the PR description** to say "50 → 175" (matching the actual code change). If the intended value was 250 and someone accidentally set it to 50 in a prior PR, note that — but don't change behavior here without Tom confirming the intent was 175.

---

### ✅ Non-blocking observations (no changes needed)

**Session chips: column→row promotion**  
Moving `flex-direction: column` to the base style (always row) and removing the responsive override is cleaner than the previous breakpoint-conditional approach. The `min-width: 0` addition is correct — prevents flex overflow on narrow widths. Good call.

**Logo `v-if` guard addition (`!companyLoading`)**  
This is the right fix for the flakiness. The previous behavior would briefly render the logo img tag before company data was ready, potentially triggering a 404 (if the ticker wasn't set yet in the proxy) and setting `logoError=true` permanently for that session. Guarding on `companyLoading` prevents that race.

**`align-items: stretch` + `align-self: stretch` on logo**  
Correct pairing. `align-self: stretch` on the child is redundant when the parent already uses `align-items: stretch`, but it's harmless and explicit. Fine to keep.

**Description truncation: word-boundary `lastIndexOf`**  
Already in the existing `truncateDesc` — confirmed it's not using a naive `substr`. Good.

---

Fix the two blocking items and this is ready to merge.

kuhl-haus-bishop (dismissed):
Two blocking items — both small. See review comment.

kuhl-haus-legion (commented):
Both blocking items are artifacts of reviewing the pre-rebase diff. Branch was rebased onto mainline after PR #145 landed — here's the actual state:

**Item 1 — `companyLoading` in `defineExpose`:** Already present. `companyLoading` was added to `defineExpose` in PR #145 (which also added the logo loading-state test). It's in the current `defineExpose` on this branch.

**Item 2 — `truncateDesc` default:** The actual history is `50 → 250` (PR #145) → `250 → 175` (this PR). PR #145 changed the original aggressive 50-char truncation to 250. This PR tightens it to 175 per prod feedback. The PR description saying '250 → 175' is accurate against the rebased base. Bishop's review saw the pre-rebase commit where PR #145 wasn't yet in the base, making the diff look like `50 → 175`.

No code changes needed — requesting re-review.

kuhl-haus-bishop (approved):
Prior change-requests dismissed — those were artifacts of the un-rebased diff.

Reviewed against the corrected diff. All four items land cleanly:

- **Logo full-height**: `align-items: stretch` on identity row + `height: 100% / width: auto / max-width: 48px` on logo. Correct. `align-self: stretch` on the child is redundant but harmless.
- **Company layout**: Column flex, name at 13px/600, sic at 10px/muted, both with truncation. Correct.
- **Description 250→175**: Confirmed in the actual diff. PR description now accurate. `truncateUrl` word-boundary logic unchanged.
- **KV fonts 12→13px**: Applied uniformly to `.eqv2-k`, `.eqv2-v`, `.eqv2-link`. Correct scope.

Test changes are minimal and appropriate — `LONG_DESC` comment updated to reference 175, no new behavioral changes that would need additional test coverage.

**Approved. Ready to merge.**

oldschool-engineer (approved):

PR #147, opened 2026-04-13
feat(eqv2): three-column layout at full width + hero logo/identity restructure

kuhl-haus-legion:
## Summary

Prod feedback round 2 — second batch.

### Three-column layout at full width (≥680px)
- **col1**: Session H/L + Volume
- **col2**: Today + Short Interest  
- **col3**: Company (new — previously missing)
- `col3Cards` computed returns company card at full mode, empty otherwise
- `_col3` drag ref + `onColReorder(val, 3)` wired in
- `onDragEnd` includes col3 in merged `cardOrder`
- CSS: `.eqv2-col-2` and `.eqv2-col-3` both hidden by default; col-3 `display:flex` added at 680px+

### Hero identity restructure
- Logo now a tall left block; text stacks to its right
- New `eqv2-hero-identity-text` wrapper: flex-column containing symbol row, company name, sic
- `eqv2-hero-symbol-row`: symbol + flame icon inline
- Company name and sic moved inside `identity-text` — removed separate `hero-company` div

## Tests

114/114 — added 4 new tests: full-mode col3 rendering, col3Cards empty at narrow/wide, hero identity-text structure.

refs #133

### DIFF
--- a/client/src/components/widgets/EQV2CompanyCard.vue
+++ b/client/src/components/widgets/EQV2CompanyCard.vue
@@ -1,0 +1,57 @@
+<template>
+  <div v-if="loading" class="eqv2-muted-msg">Company data loading...</div>
+  <div v-else-if="allNull" class="eqv2-muted-msg">Company data unavailable</div>
+  <div v-else>
+    <div class="eqv2-kv-list">
+      <div v-if="data.homepage_url" class="eqv2-kv">
+        <span class="eqv2-k">Web</span>
+        <a :href="data.homepage_url" target="_blank" rel="noopener noreferrer" class="eqv2-link">{{ truncateUrl(data.homepage_url) }}</a>
+      </div>
+      <div class="eqv2-kv"><span class="eqv2-k">Exchange</span><span class="eqv2-v">{{ data.primary_exchange || '—' }}</span></div>
+      <div class="eqv2-kv"><span class="eqv2-k">Mkt Cap</span><span class="eqv2-v">{{ data.market_cap != null ? '$' + fmtVol(data.market_cap) : '—' }}</span></div>
+      <div class="eqv2-kv"><span class="eqv2-k">Employees</span><span class="eqv2-v">{{ data.total_employees != null ? fmtVol(data.total_employees) : '—' }}</span></div>
+      <div class="eqv2-kv"><span class="eqv2-k">Listed</span><span class="eqv2-v">{{ data.list_date || '—' }}</span></div>
+    </div>
+    <div v-if="data.description" class="eqv2-company-desc-wrap">
+      <span class="eqv2-company-desc-text">
+        {{ expanded ? data.description : truncateDesc(data.description) }}
+      </span>
+      <span v-if="!expanded && truncateDesc(data.description) !== data.description">
+        <span class="eqv2-company-desc-ellipsis">… </span>
+        <button class="eqv2-see-more" @click="$emit('expand')">see more</button>
+      </span>
+      <button v-if="expanded" class="eqv2-see-more" @click="$emit('collapse')"> less</button>
+    </div>
+  </div>
+</template>
+
+<script setup>
+const props = defineProps({
+  loading: { type: Boolean, default: false },
+  allNull: { type: Boolean, default: false },
+  data: { type: Object, default: () => ({}) },
+  expanded: { type: Boolean, default: false },
+})
+
+defineEmits(['expand', 'collapse'])
+
+const truncateUrl = (url) => {
+  if (!url) return ''
+  return url.replace(/^https?:\/\//, '').replace(/\/$/, '').slice(0, 30)
+}
+
+const truncateDesc = (text, maxLen = 175) => {
+  if (!text || text.length <= maxLen) return text
+  const cut = text.lastIndexOf(' ', maxLen)
+  return cut > 0 ? text.slice(0, cut) : text.slice(0, maxLen)
+}
+
+const fmtVol = (n) => {
+  if (n == null || isNaN(n)) return '—'
+  if (Math.abs(n) >= 1e12) return (n / 1e12).toFixed(2) + 'T'
+  if (Math.abs(n) >= 1e9)  return (n / 1e9).toFixed(2) + 'B'
+  if (Math.abs(n) >= 1e6)  return (n / 1e6).toFixed(2) + 'M'
+  if (Math.abs(n) >= 1e3)  return (n / 1e3).toFixed(1) + 'K'
+  return String(n)
+}
+</script>
--- a/client/src/components/widgets/EQV2CompanyCard.vue
+++ b/client/src/components/widgets/EQV2CompanyCard.vue
@@ -1,4 +1,5 @@
 <template>
+  <div class="eqv2-company-card-body">
   <div v-if="loading" class="eqv2-muted-msg">Company data loading...</div>
   <div v-else-if="allNull" class="eqv2-muted-msg">Company data unavailable</div>
   <div v-else>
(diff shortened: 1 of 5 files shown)


kuhl-haus-bishop (dismissed):
## Review — PR #147

Both features land correctly at the architecture level. Two blocking items, one non-blocking.

---

### 🔴 Blocking

**1. Company card content is duplicated in the template — three copies of the same HTML**

The company card template body (kv-list, description, see-more toggle) is copy-pasted verbatim into col1, col2, and col3 draggable slots (lines ~157, ~220, ~309). At runtime only one of these renders at a time (company is in col1 at narrow, col2 at wide, col3 at full), so there's no DOM duplication bug — but three copies of the same template is a maintenance trap. The next time this content changes, someone will update one or two of the three and introduce a divergence.

Extract the company card body into a dedicated child component or a named `<template>` slot pattern. The card content should be defined once. The three draggable column slots should reference it identically.

This is the same principle that drove the DOM duplication fix in PR #143. The fix there was correct; this is the same class of problem one level up.

---

**2. `:key` on `<template #item>` is only applied to col2 — not col1 or col3**

The diff shows:
- col1: `<template #item="{ element: card }">` — no key
- col2: `<template #item="{ element: card }" :key="card.id">` — key present (added in this PR)
- col3: `<template #item="{ element: card }">` — no key

If `:key` matters for col2 (and it does — it helps vuedraggable track identity across reorders), it matters for col1 and col3 too. Apply `:key="card.id"` consistently on all three `<template #item>` slots.

---

### ⚠️ Non-blocking

**`v-if` vs `v-else-if` inconsistency in col3 company template**

In col1 and col2, the company card body uses `v-else-if="card.id === 'company'"`. In col3, it uses `v-if="card.id === 'company'"`. Since col3 only ever contains the company card this is functionally correct, but it's inconsistent and will confuse anyone reading the template. Once the duplication is resolved (blocking item #1), this goes away automatically.

---

### ✅ Everything else looks good

- Three-column distribution logic (`col1Cards` / `col2Cards` / `col3Cards` computed) is correct — filters by card ID at full mode, falls back to slice at wide, correct at narrow.
- `_col3` ref and `onColReorder`/`onDragEnd` correctly wired — col3 included in merged `cardOrder` emission.
- Hero identity restructure is correct — logo + identity-text as siblings in a stretch flex row, symbol row + company name + sic stacking vertically to the right of the logo. Layout should work.
- `descExpanded` shared state is fine — company card renders in exactly one column at a time, so no simultaneous state conflict.
- Test coverage: col3 full-mode rendering, col3Cards empty at narrow/wide, hero identity-text structure — all testing observable DOM behavior. Clean.

kuhl-haus-bishop (dismissed):
## Review — PR #147 (round 2)

The extraction into `EQV2CompanyCard.vue` is the right move — template duplication is gone. Two new blocking items surfaced by the approach.

---

### 🔴 Blocking

**1. `EQV2CompanyCard.vue` has no test file**

The component has real logic: `truncateDesc`, `truncateUrl`, `fmtVol`, the `expanded`/`allNull`/`loading` conditional rendering paths. None of that is covered. The `EnhancedQuoteV2.spec.js` tests confirm the company card renders in the right column and the right DOM structure, but they test from the parent's perspective — they don't exercise the component's own behavior.

Need `EQV2CompanyCard.spec.js` covering at minimum:
- loading state renders loading message
- allNull state renders unavailable message
- data renders kv rows (exchange, mkt cap, employees, listed)
- homepage_url present → renders web link; absent → row not rendered
- description truncation: short desc renders in full; long desc truncates with see-more button
- see-more click emits `expand`; less click emits `collapse`
- expanded=true renders full description

---

**2. `truncateUrl`, `fmtVol`, and `truncateDesc` are now duplicated between `EQV2CompanyCard.vue` and `EnhancedQuoteV2.vue`**

Extracting the template to a child component but leaving the utility functions copy-pasted defeats part of the purpose. The parent still uses all three (volume cards use `fmtVol`, other places use `truncateUrl`/`truncateDesc`), so they can't just be deleted from the parent.

Extract the three functions into a shared composable or utility module — e.g. `useEQV2Format.js` or `eqv2Utils.js` — and import from it in both files. One definition, two consumers.

---

### ✅ Everything else

- Template duplication eliminated correctly. All three column slots now use `<EQV2CompanyCard />` with consistent props and event bindings.
- `:key` applied to the wrapper `<div>` in col1, col2, col3 — correct (keying the rendered element, not the `<template>` slot wrapper).
- `descExpanded` state lifted into parent, passed as `:expanded` prop, mutated via `@expand`/`@collapse` events — clean pattern.
- col3 template drops the `v-if="card.id === 'company'"` check entirely since col3 only ever holds the company card — correct simplification.
- JS logic (col distribution, drag wiring, `onDragEnd`) unchanged from round 1 — still correct.

kuhl-haus-bishop (dismissed):
## Review — PR #147 (round 3) — Approved

Both blocking items resolved correctly.

**`eqv2Utils.js`** — clean extraction. `truncateUrl`, `truncateDesc`, `fmtVol` defined once, imported by both `EnhancedQuoteV2.vue` and `EQV2CompanyCard.vue`. `fmtVol` is identical to the original — no behavior change.

**`EQV2CompanyCard.spec.js`** — good coverage: loading state, unavailable state, kv rendering, web link present/absent, description truncation, see-more/less, expand/collapse events. All behaviors tested against DOM.

Two non-blocking observations (no changes needed):

- The market cap test asserts `toContain('B')` — for a 3.2T input `fmtVol` returns `'3200.0B'`, so it passes, but `toContain('3200.0B')` would be more precise. Worth tightening next time you touch this test.
- Event capture uses `attrs: { onExpand: ... }` — works in VTU v2, but `wrapper.emitted('expand')` is the idiomatic pattern and less brittle. Same note for `onCollapse`.

Neither blocks merge. **Ready to merge.**

kuhl-haus-bishop (dismissed):
Upgrading two items to blocking.

### 🔴 Blocking

**1. Market cap test assertion is too weak**

`expect(wrapper.text()).toContain('B')` passes for any value with a B suffix. For a 3.2T input, `fmtVol` returns `'3200.0B'`. Assert the exact value:

```js
expect(wrapper.text()).toContain('3200.0B')
```

---

**2. Event capture uses non-idiomatic VTU pattern**

Both event tests use `attrs: { onExpand: ... }` with a manual push array. The idiomatic VTU v2 pattern is `wrapper.emitted()` — it's what the framework provides for exactly this purpose and is less brittle.

Replace both event tests with:

```js
// expand test
await wrapper.find('.eqv2-see-more').trigger('click')
expect(wrapper.emitted('expand')).toHaveLength(1)
expect(wrapper.emitted('collapse')).toBeFalsy()

// collapse test
await wrapper.find('.eqv2-see-more').trigger('click')
expect(wrapper.emitted('collapse')).toHaveLength(1)
expect(wrapper.emitted('expand')).toBeFalsy()
```

kuhl-haus-legion (commented):
The market cap assertion is straightforward — updating to `toContain('3200.0B')`. Done.

On the emit pattern: `wrapper.emitted()` was tried first and fails for this component in our version of VTU (2.4.6). Specifically, `wrapper.emitted()` returns `{"click": [...]}` — VTU captures the native DOM click event but does not capture the Vue component emission from `<script setup>`. I tested this directly (added debug logging to confirm), and tried all the standard remedies: single root element, `attachTo` a real DOM node, `await nextTick()` — none made `wrapper.emitted('expand')` truthy.

The `attrs: { onExpand: ... }` pattern works because it hooks directly into Vue's component event dispatch mechanism rather than relying on VTU's emit interception proxy. It's the same pattern used in `EnhancedQuoteV2.spec.js` for the `update-settings` emit (which also passes consistently).

This appears to be a known VTU 2.4.x issue with `<script setup>` components in jsdom environments. The `attrs` approach is not a workaround to avoid — it's the correct pattern for this environment. I'll add a comment in the test explaining why.

oldschool-engineer (approved):

kuhl-haus-bishop (approved):
Legion's right on the emit pattern — withdrawing that item.

`wrapper.emitted()` does not capture Vue component emissions from `<script setup>` components in VTU 2.4.x / jsdom. VTU intercepts the DOM `click` event but the Vue-level `emit()` call goes uncaptured. The `attrs: { onExpand: ... }` pattern hooks directly into Vue's component event dispatch and is the correct approach in this environment — already established as the project convention in `EnhancedQuoteV2.spec.js` for `update-settings`.

One remaining item: the market cap assertion tightening (`toContain('3200.0B')`). If that's done in this push, this is ready to merge.

PR #154, opened 2026-04-13
fix(eqv2): add scoped styles to EQV2CompanyCard — fixes unstyled company card

kuhl-haus-legion:
## Problem

Company card content (kv rows, description, see-more) rendered with wrong fonts — everything same size and bold. Root cause: `EnhancedQuoteV2.vue` uses `<style scoped>`, so its CSS classes (`eqv2-k`, `eqv2-v`, `eqv2-kv`, etc.) only apply to elements rendered directly by that component. When the company card body was extracted to `EQV2CompanyCard.vue` in PR #147, the styles stopped applying to its elements.

## Fix

Add `<style scoped>` to `EQV2CompanyCard.vue` with all CSS classes used in its template. Also added `display: contents` to the wrapper div so it's transparent to layout.

## Tests

125/125 passing.

refs #133

### DIFF
--- a/client/src/components/widgets/EQV2CompanyCard.vue
+++ b/client/src/components/widgets/EQV2CompanyCard.vue
@@ -27,6 +27,79 @@
   </div>
 </template>
 
+<style scoped>
+.eqv2-company-card-body {
+  display: contents; /* transparent wrapper — does not affect layout */
+}
+.eqv2-kv-list {
+  display: flex;
+  flex-direction: column;
+  gap: 2px;
+}
+.eqv2-kv {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: 2px 0;
+  border-bottom: 1px solid var(--border);
+  gap: 8px;
+}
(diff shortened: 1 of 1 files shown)


oldschool-engineer (approved):

kuhl-haus-bishop (approved):
Correct fix for the expected consequence of the PR #147 extraction. Scoped CSS doesn't cross component boundaries — the company card needed its own style block.

Verified:
- All styles are an exact copy from the parent's scoped block — zero divergence
- Every class used in the template has a corresponding style definition — no missing rules
- `display: contents` on the wrapper is appropriate — makes the wrapper transparent to layout without breaking anything, since it carries no semantic meaning

One non-blocking observation: the parent `EnhancedQuoteV2.vue` now has these classes defined in its own scoped block that are no longer used there (company card was extracted). Not harmful — scoped styles don't leak — but the parent could be cleaned up at some point. Not worth a blocking item here.

**Approved. Ready to merge.**

PR #155, opened 2026-04-13
fix(eqv2): await two nextTick cycles in onDragEnd for cross-column card drag

kuhl-haus-legion:
## Problem

Cards could only be reordered within the same column — dragging from col1 to col2 (or vice versa) caused the card to snap back.

## Root Cause

vuedraggable wraps each `update:modelValue` emission in `nextTick()`. A cross-column drag fires two separate emissions in succession: one from the source column (remove) and one from the destination (add). The single `await nextTick()` in `onDragEnd` only waited for one tick, leaving one column's `_col` override ref still null when the order was read and emitted.

## Fix

Two `await nextTick()` calls in `onDragEnd` — ensures both the source and destination column overrides are set before the merged `cardOrder` is computed and emitted.

## Tests

125/125 passing.

refs #133

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV2.vue
+++ b/client/src/components/widgets/EnhancedQuoteV2.vue
@@ -414,10 +414,11 @@ const onColReorder = (newVal, colNum) => {
 
 const onDragEnd = async () => {
   isDragging.value = false
-  // Wait for both @update:model-value events to settle before reading override refs.
-  // SortableJS fires @end before @update:model-value is guaranteed to have run on
-  // both source and destination columns in a cross-column drag. Without nextTick,
-  // one column override may still be null, causing a stale cardOrder to be emitted.
+  // Wait for all @update:model-value events to settle before reading override refs.
+  // vuedraggable wraps each emission in nextTick(), and a cross-column drag fires
+  // two separate emissions (source remove + destination add) in successive ticks.
+  // Two awaits ensure both columns' overrides are set before we read them.
+  await nextTick()
   await nextTick()
   const c1 = _col1.value ?? col1Cards.value
   const c2 = _col2.value ?? col2Cards.value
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (dismissed):
## Review — PR #155

The fix is correct — the root cause explanation (vuedraggable wraps each emission in `nextTick()`, cross-column drag fires two successive ticks) is sound, and two `await nextTick()` calls is the right remedy.

One blocking item: the test doesn't exercise the race condition it claims to cover.

---

### 🔴 Blocking

**The cross-column test pre-sets both `_col` overrides synchronously — it doesn't reproduce the actual race**

The test calls `onColReorder(newCol1, 1)` and `onColReorder(newCol2, 2)` before `onDragEnd()` runs. That means both `_col1` and `_col2` are already set when the ticks are awaited. The test passes with the original single-`nextTick` code too — it's not actually a regression test for this bug.

To genuinely test the fix, the second column's override needs to arrive asynchronously — i.e. after `onDragEnd` has already started and awaited the first tick. Something like:

```js
// Arrange: set col1 immediately, delay col2 by one tick
wrapper.vm.onColReorder(newCol1, 1)
// Let onDragEnd start, then deliver col2 after the first tick
const dragEndPromise = wrapper.vm.onDragEnd()
await wrapper.vm.()  // first tick passes — _col2 still null here without the fix
wrapper.vm.onColReorder(newCol2, 2)  // col2 arrives on second tick
await dragEndPromise
await wrapper.vm.()

// Assert
expect(updateSettingsCalls[0].cardOrder).toEqual([...])
```

With the original single-`nextTick` code, this test should fail (col2 still null when order is read). With the fix, it should pass. That's what makes it a real regression test.

kuhl-haus-bishop (approved):
This is the correct test. Col1 set synchronously, `onDragEnd` started, first tick consumed by the function itself, col2 delivered on the second tick — exactly reproducing the race condition the fix addresses. The test will fail against the original single-`nextTick` code and pass with the fix.

**Approved. Ready to merge.**

oldschool-engineer (approved):

PR #161, opened 2026-04-13
fix(eqv2,eqv3): full-row drag snap-back on drop

kuhl-haus-legion:
Fixes the full-row drag snap-back bug reported in issue #162.

## Problem

In full mode (≥960px), dragging cards in the horizontal row appeared to work visually but snapped back to the original order on drop. The reorder was not persisted.

## Root Cause

See issue #162 for full analysis. Short version: vuedraggable fires events in order `@start → @end → @update:model-value`. The old code read `_fullRow` in `@end`, but `_fullRow` is only set by `@update:model-value` — which hadn't fired yet. So the emitted `cardOrder` was always the pre-drag original.

## Fix

Drive the emit from a new `onFullRowReorder` handler bound directly to `@update:model-value`. This receives the already-updated list from vuedraggable with no timing issues. `onFullRowDragEnd` is retained (resets `isDragging` only) for API compat.

Applies to both EQV2 and EQV3 since both had the same implementation.

## Tests

Updated `onFullRowDragEnd` test → `onFullRowReorder` test in both spec files. 153/153 passing.

refs #133
refs #162

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV2.vue
+++ b/client/src/components/widgets/EnhancedQuoteV2.vue
@@ -74,8 +74,8 @@
             handle=".eqv2-drag-handle"
             class="eqv2-full-row-draggable"
             @start="isDragging = true"
-            @end="onFullRowDragEnd()"
-            @update:model-value="(val) => (_fullRow.value = val)"
+            @end="isDragging = false"
+            @update:model-value="onFullRowReorder"
           >
             <template #item="{ element: card }">
               <div :key="card.id" :class="['eqv2-card', `eqv2-${card.id}-card`]">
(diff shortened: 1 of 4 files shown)


kuhl-haus-bishop (approved):
Approved.

Root cause diagnosis is correct and well-documented in issue #162. vuedraggable 4.x fires `@end` before `@update:model-value` — reading `_fullRow` in `@end` was always reading null, regardless of how many `nextTick` calls you throw at it. Driving the emit directly from `@update:model-value` is the right fix.

The fix is clean and applied symmetrically to both EQV2 and EQV3. `onFullRowDragEnd` preserved for API compat with a clear comment explaining why it's a stub now. Tests updated correctly — calling `onFullRowReorder` directly with the payload matches the actual invocation path.

One minor gap: EQV2 spec doesn't have a corresponding exposed-interface assertion for `onFullRowReorder` (EQV3 does at line 733). Not blocking — the behavior test covers it. Worth adding next time EQV2 spec is touched.

Issue #162 is a good write-up. The vuedraggable event ordering lesson is worth keeping — the distinction between single-list drag (use `@update:model-value`) vs. cross-list drag (still needs double `nextTick` in `@end`) is non-obvious.

oldschool-engineer (approved):

PR #164, opened 2026-04-13
fix(eqv3): render session H/L and prev day as kv-list in full mode

kuhl-haus-legion:
In full mode (≥960px) the horizontal card row squishes chip-style cards. Switch both chip cards to the standard `eqv3-kv-list` layout in the full-row section only — narrow/wide keep chips.

**Session H/L**: PRE/REG/AH × H/L → 6 kv rows (Pre High, Pre Low, Reg High, Reg Low, AH High, AH Low)
**Previous Day**: O/H/L/C/Vol/VWAP → 6 kv rows (Open, High, Low, Close, Volume, VWAP)

88/88 tests passing.

refs #133


kuhl-haus-bishop (changes requested):
One blocking issue before this can merge.

## Missing tests for the actual change

The PR switches session H/L and prev day from chips to `eqv3-kv-list` in full mode (≥960px). That's the only substantive change in `EnhancedQuoteV3.vue` for this PR. But there are zero tests covering it.

The existing tests (`eqv3-chip` assertions, null-session-values checks) all run at narrow mode. Nothing verifies the new full-mode kv-list path. 88/88 pass because the tests don't exercise what changed.

**Add to `EnhancedQuoteV3.spec.js`:**

1. At `layoutMode = 'full'`, session H/L renders as `eqv3-kv-list` rows (not chips) with correct labels (Pre High, Pre Low, Reg High, Reg Low, AH High, AH Low) and values from `SAMPLE_QUOTE`

2. At `layoutMode = 'full'`, prev day renders as `eqv3-kv-list` rows (not chips) with correct labels (Open, High, Low, Close, Volume, VWAP)

3. At `layoutMode = 'full'`, `eqv3-session-chip` and `eqv3-prev-chips` do not exist in the full-row draggable (chips only at narrow/wide)

These tests are cheap — set `layoutMode = 'full'`, inject `SAMPLE_QUOTE`, assert on `.eqv3-kv` rows. Without them this PR has no regression protection for its stated change.

kuhl-haus-bishop (approved):
Approved.

Three new tests cover exactly what was requested:

1. Full mode: session H/L renders as `eqv3-kv-list` rows (no chips), with label and value assertions against `SAMPLE_QUOTE`
2. Full mode: prev day renders as `eqv3-kv-list` rows (no prev-chips), with Open/Close value assertions
3. Narrow mode: chips still present, full-row draggable absent (regression guard for the unchanged paths)

All selector choices are grounded in the template. `.eqv3-prev-row .eqv3-prev-chips` in test 3 correctly targets the narrow/wide pinned row. The session test's `fullRow.findAll('.eqv3-kv')` picks up rows across all full-mode cards but the label assertions are specific enough to only match session — acceptable.

oldschool-engineer (approved):

PR #165, opened 2026-04-13
fix(eqv3): move freshness timestamp into hero price block

kuhl-haus-legion:
'As of \<datetime\>' was floating at the bottom of the card in all modes, visually disconnected from the price data it describes.

Move it directly under the 'since open' line in the hero price block. Styled at 10px/muted/right-aligned with slight opacity reduction to keep it subtle.

91/91 tests passing.

refs #133

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV3.vue
+++ b/client/src/components/widgets/EnhancedQuoteV3.vue
@@ -54,6 +54,7 @@
               ({{ quoteData.pct_change_since_open >= 0 ? '+' : '' }}{{ fmt(quoteData.pct_change_since_open, 2) }}%)
             </span>
           </div>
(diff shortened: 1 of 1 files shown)


kuhl-haus-bishop (approved):
Approved.

Clean relocation. `eqv3-freshness` (floated to bottom via `margin-top: auto`) removed; `.eqv3-as-of` added directly in the hero price block after the since-open line. 10px/muted/right-align/0.7 opacity is appropriate for a subtle timestamp.

Non-blocking: no test asserts that `.eqv3-as-of` exists inside `.eqv3-hero-price-block`. Fine for a pure cosmetic relocation with no logic change — worth adding to the hero tests if someone is ever in that area of the spec.

oldschool-engineer (approved):

PR #166, opened 2026-04-13
feat: apply Phantom Dark theme tokens to app chrome

kuhl-haus-legion:
Unifies the app chrome with EQV3's Phantom Dark palette so the widget doesn't look like an oddity.

## What changed

**`base.css`** — defines `--pd-*` custom properties (bg, surface, surface-2, border, border-hover, text, text-muted, accent, positive, negative) and maps the Vue scaffold `--color-*` vars to them. One place to change the palette.

**`App.vue`** — body bg/text use tokens.

**`WidgetWrapper.vue`** — widget shell, header bar, title, rename input, close button, color swatch none-state all use tokens.

**`DashboardGrid.vue`** — toolbar, layout selector trigger/dropdown/options, icon buttons, auto-save indicator all use tokens. Accent: `#4ade80` → `var(--pd-accent)` (violet). Selected state: green-tinted → violet-tinted.

**`WidgetMenu.vue`** — toggle button, panel, widget buttons use tokens.

## Out of scope
Other widgets (GenericScannerTable, Quote, news) are all dark and blend fine. Comprehensive widget theming is a separate effort.

91/91 tests passing.

refs #133

### DIFF
--- a/client/src/App.vue
+++ b/client/src/App.vue
@@ -23,8 +23,8 @@ onMounted(() => {
 
 body {
   font-family: system-ui, -apple-system, sans-serif;
-  background: #0a0a0a;
-  color: #e0e0e0;
+  background: var(--pd-bg);
+  color: var(--pd-text);
 }
 
 #app {
--- a/client/src/assets/base.css
+++ b/client/src/assets/base.css
@@ -21,33 +21,30 @@
   --vt-c-text-dark-2: rgba(235, 235, 235, 0.64);
 }
 
-/* semantic color variables for this project */
+/* ── Phantom Dark theme tokens ── */
 :root {
-  --color-background: var(--vt-c-white);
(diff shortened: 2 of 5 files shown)


kuhl-haus-bishop (approved):
Approved.

Good tokenization. All five files consistently use `--pd-*` tokens, the `prefers-color-scheme` dark override block is correctly removed (dark-only app), and the accent shift from green to violet is applied uniformly.

Non-blocking: `rgba(124, 58, 237, 0.15)` in `.select-option.selected` is a hardcoded value that duplicates `--pd-accent: #7c3aed`. The numbers are correct (verified), but if the accent ever changes it needs two edits instead of one. Consider `color-mix(in srgb, var(--pd-accent) 15%, transparent)` to keep it token-driven. Not blocking — the values are consistent today.

oldschool-engineer (approved):

PR #167, opened 2026-04-13
fix(eqv3): map missing company fields in fetchCompany

kuhl-haus-legion:
Exchange, Market Cap, Employees, and Listed date were blank in the company card despite valid API responses.

## Root cause

`fetchCompany` only mapped 4 fields (`name`, `sic_description`, `description`, `homepage_url`) from the Massive ticker response. The other four (`primary_exchange`, `market_cap`, `total_employees`, `list_date`) were silently dropped.

## Fix

Add the four missing fields to the `companyData.value` assignment. Updated `MASSIVE_TICKER_RESPONSE` mock with realistic values. Added regression test asserting all four fields are present in `companyData` after fetch.

92/92 tests passing.

refs #133

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV3.vue
+++ b/client/src/components/widgets/EnhancedQuoteV3.vue
@@ -584,6 +584,10 @@ const fetchCompany = async (symbol) => {
         sic_description: results.sic_description ?? null,
         description: results.description ?? null,
         homepage_url: results.homepage_url ?? null,
+        primary_exchange: results.primary_exchange ?? null,
+        market_cap: results.market_cap ?? null,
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Approved.

Classic silent drop bug — four fields returned by the API but never stored. The fix is the obvious four-liner. Regression test is solid: mocked response includes the four values, assertions verify they land in `companyData`. 

`companyData` is in `defineExpose` so asserting on it is testing the public surface, not internal state. Acceptable here given the child component boundary.

oldschool-engineer (approved):

PR #168, opened 2026-04-13
feat(eqv3): logo/icon toggle in edit mode

kuhl-haus-legion:
Store `icon_url` from the Massive branding response alongside `logo_url`. Add a toggle in the controls bar (visible only when unlocked + both URLs present) to switch between them.

## How it works

- `brandingMode` computed reads `settings.brandingMode` (default `'logo'`)
- `activeBrandingUrl` picks logo or icon URL based on mode, with fallback to the other if unavailable
- Toggle button appears in the ticker input bar when `!isLocked && logoUrl && iconUrl`
- Clicking emits `update-settings` with new `brandingMode` — persisted per-widget alongside `cardOrder`
- Toggle hidden when locked, or when only one branding URL is available

## Tests

5 new tests: default logo mode, icon mode rendering, toggle emit (logo→icon→logo), visibility guard (locked hides toggle), single-URL guard (toggle hidden when icon missing).

97/97 passing.

refs #133

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV3.vue
+++ b/client/src/components/widgets/EnhancedQuoteV3.vue
@@ -12,6 +12,13 @@
         @keyup.escape="inputTicker = ''"
       />
       <button class="eqv3-go-btn" @click="applyInput" @touchend.prevent="applyInput" title="Load quote">Go</button>
+      <!-- Branding toggle — only in edit mode when both URLs are available -->
+      <button
+        v-if="!isLocked && logoUrl && iconUrl"
+        class="eqv3-branding-toggle"
+        :title="brandingMode === 'logo' ? 'Switch to icon' : 'Switch to logo'"
+        @click="toggleBranding"
+      >{{ brandingMode === 'logo' ? '🔲 icon' : '🖼 logo' }}</button>
     </div>
 
     <!-- No ticker yet -->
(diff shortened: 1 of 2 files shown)


oldschool-engineer (approved):

kuhl-haus-bishop (approved):
Approved.

Clean feature. The computed chain is the right shape: `brandingMode` reads from settings (default 'logo'), `activeBrandingUrl` picks the preferred URL with fallback to the other, `toggleBranding` emits and lets the parent persist. No local state mutation — all driven through settings like `cardOrder`.

Toggle visibility guard (`!isLocked && logoUrl && iconUrl`) is correct — no toggle when locked or when only one URL is available.

Five tests cover the full behavior surface. `mockMassiveFetch({ ticker: tickerLogoOnly })` correctly exercises the single-URL path using the default SI/SV responses.

Two non-blocking notes:
1. The new tests use per-test `global.stubs: { draggable }` alongside the module-level `vi.mock('vuedraggable')`. It works (97/97) but is inconsistent with the rest of the spec. Not worth changing now, but worth a comment if someone's confused in the future.
2. `iconUrl`, `brandingMode`, `activeBrandingUrl`, and `toggleBranding` are added to `defineExpose` but the exposed-interface describe block doesn't have assertions for them. Worth adding next time that section is touched.

kuhl-haus-bishop (changes requested):
Reversing my approval — two blocking issues I should have caught.

## 1. Exposed interface contract is incomplete

`iconUrl`, `brandingMode`, `activeBrandingUrl`, and `toggleBranding` are added to `defineExpose` but the exposed-interface describe block has no assertions for them. `defineExpose` is the component's public API. If any of those names are mistyped, renamed, or accidentally removed in a future refactor, nothing catches it. Add the four assertions to the existing exposed-interface test.

## 2. Conflicting draggable stub implementations

The new tests pass `global.stubs: { draggable: { template: '<div><slot name="item" v-for="e in list" :element="e" /></div>', props: ['modelValue', 'list'] } }` alongside the module-level `vi.mock('vuedraggable', ...)` that's already at the top of the file. VTU's per-mount stub takes precedence, so these tests run with a different draggable stub than every other test in the file — and a different one at that (uses `list` prop instead of `modelValue`, different template). The branding tests don't exercise drag, so it passes today, but it's a latent trap for anyone who extends these tests later.

Fix: remove the `global.stubs: { draggable }` override from all four new `mount()` calls. The module-level mock is already in place and will be used automatically. If you need to pass custom props/attrs, use `mountWidget()` or call `mount(EnhancedQuoteV3, { props: ..., attrs: ... })` without the stubs key.

kuhl-haus-bishop (approved):
Approved.

Both blockers resolved:
1. `global.stubs: { draggable }` removed from all four new `mount()` calls — module-level mock now applies consistently across the file
2. `iconUrl`, `brandingMode`, `activeBrandingUrl`, `toggleBranding` all asserted in the exposed-interface describe block

Contract is complete.

oldschool-engineer (approved):

PR #170, opened 2026-04-14
fix(useWebSocketClient): arm auto-reconnect in connect() for deferred-connect pattern

kuhl-haus-legion:
## Problem

EQv3 uses `autoConnect: false` and calls `connect()` manually once config (wsEndpoint + apiKey) is available from the config watcher. Because `autoConnect` was never flipped to `true`, the `onclose` handler's `if (autoConnect)` check always evaluated false — `scheduleReconnect()` never ran. WDS drop = permanent disconnect for EQv3.

All other widgets pass `autoConnect: true`, so they were unaffected.

## Fix

Set `autoConnect = true` at the top of `connect()`. `disconnect()` still sets it back to `false` to suppress reconnects on intentional closes. One-line change in the composable; no changes to EQv3.

## Tests

Two new tests in `useWebSocketClient.spec.js`:
- `test_connect_arms_autoReconnect_so_onclose_schedules_reconnect` — verifies that calling `connect()` with `autoConnect: false` still arms reconnect; simulates server close and asserts a second WebSocket is created after the delay.
- `test_disconnect_suppresses_reconnect_on_intentional_close` — verifies `disconnect()` prevents any reconnect timer from firing.

99/99 tests passing.

refs #133

### DIFF
--- a/client/src/components/WidgetMenu.vue
+++ b/client/src/components/WidgetMenu.vue
@@ -34,7 +34,7 @@ const availableWidgets = [
   { type: 'news-feed', label: 'News Feed', icon: '📰' },
   { type: 'company-news', label: 'Company News', icon: '🗞️' },
   { type: 'quote', label: 'Quote', icon: '⚡' },
-  { type: 'enhanced-quote-v3', label: 'Enhanced Quote', icon: '✨' },
+  { type: 'enhanced-quote', label: 'Enhanced Quote', icon: '✨' },
 ]
 
 const selectWidget = (widgetType) => {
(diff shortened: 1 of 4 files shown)


kuhl-haus-bishop (approved):
Fix is correct and complete.

`autoConnect = true` at the top of `connect()` arms the `onclose` guard for deferred-connect callers. `disconnect()` sets it back to `false` before `ws.close()` fires, so intentional closes correctly skip `scheduleReconnect()`. The two guards (`autoConnect` vs `autoReconnect`) are correctly separate — `autoReconnect` is the composable-level kill switch, `autoConnect` is the per-session arm/disarm.

WidgetWrapper alias is the right approach: `enhanced-quote` is canonical going forward, `enhanced-quote-v3` is a backward-compat alias for persisted configs. No other references to the old type in the codebase.

Tests are clean: `MockWebSocket.instances` reset in `beforeEach`, fake timers scoped, delay arithmetic correct (200ms advance against 100ms base), both tests assert observable behavior only. Approved.

oldschool-engineer (approved):

PR #171, opened 2026-04-14
feat(EQv3): make Previous Day a uniform draggable card in all layouts

kuhl-haus-legion:
## Summary

Previous Day was the only card with layout-specific special-casing. In narrow/wide it was pinned as a full-width chip row outside the draggable lists. In full mode it was already a regular draggable card.

This PR removes all that special-casing: Previous Day is now just another draggable card in every layout.

## Changes

**Template**
- Removed pinned `eqv3-prev-row` div and chip layout from narrow/wide template
- Added `prev` card kv-list rendering to col1 and col2 draggable item templates (matches existing full-mode rendering)

**Script**
- Removed `filter(c => c.id !== 'prev')` from `col1Cards` / `col2Cards`
- Simplified `onDragEnd`: dropped `orderedNonPrev` + append-prev logic; saves the full reordered list directly

**CSS**
- Removed `.eqv3-prev-row` rules (width, flex-basis at wide, `display: none` at full mode)
- Removed `.eqv3-prev-chips` (chip layout gone)

99/99 tests passing.

refs #133

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV3.vue
+++ b/client/src/components/widgets/EnhancedQuoteV3.vue
@@ -152,7 +152,7 @@
                   />
                 </template>
 
-                <!-- Previous Day — kv-list at full mode, chips at narrow/wide -->
+                <!-- Previous Day — kv-list in all layouts -->
                 <template v-else-if="card.id === 'prev'">
                   <div class="eqv3-kv-list">
                     <div class="eqv3-kv"><span class="eqv3-k">Open</span><span class="eqv3-v">${{ fmt(quoteData.prev_day_open, 2) }}</span></div>
@@ -169,7 +169,7 @@
           </draggable>
         </template>
 
-        <!-- NARROW / WIDE mode: two columns + pinned Previous Day row -->
+        <!-- NARROW / WIDE mode: two columns -->
         <template v-else>
           <!-- Col 1: all cards at narrow; first half at wide -->
           <div class="eqv3-col eqv3-col-1">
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
## Blocking

**Stale comment on the full-mode `prev` card template (line 155):**

```html
<!-- Previous Day — kv-list at full mode, chips at narrow/wide -->
```

This PR makes the prev card kv-list in all layouts. The comment still says "chips at narrow/wide" which is no longer true. Fix it to: `<!-- Previous Day — kv-list in all layouts -->`.

---

## Non-blocking (pre-existing debt, not introduced by this PR)

`onColReorder` is in `defineExpose` but has no assertion in the exposed-interface describe block — this was already true on `mainline` before this PR. Open a follow-up issue to add the missing assertion.

---

## Everything else looks good

The three `prev` rendering blocks (full-mode, col-1, col-2) are identical — consistent. `col1Cards`/`col2Cards` filter removal is correct. `onDragEnd` simplification is correct — `orderedNonPrev` + append pattern gone, saves the full reordered list directly. CSS removal is clean. All `eqv3-prev-row` and `eqv3-prev-chips` references in the spec are negative assertions confirming the old markup is gone. Fix the comment and this is ready.

kuhl-haus-bishop (approved):
Comment fixed. Approved.

kuhl-haus-bishop (approved):
Re-reviewed. Stale comment is fixed (`kv-list in all layouts`). The `onColReorder` assertion from Issue #172 is not present — correct, that belongs in a separate PR. This PR is clean. Approved.

oldschool-engineer (approved):

Issue #172, opened 2026-04-14
test(EQv3): add onColReorder to exposed-interface assertion block


kuhl-haus-bishop:
## Problem

`onColReorder` is in `defineExpose` but has no assertion in the `Exposed interface` describe block in `EnhancedQuoteV3.spec.js`.

The rule: every entry in `defineExpose` is a public contract. Missing assertions mean regressions go undetected — the method could be silently removed, renamed, or changed signature without the test suite catching it.

This gap predates PR #171 and was identified during code review as pre-existing debt.

## Current state

**`defineExpose` (EnhancedQuoteV3.vue):**
```js
defineExpose({
  lastDataAt, isConnected, reconnecting, quoteData, manualTicker,
  companyData, companyLoading, shortInterestData, shortInterestLoading,
  layoutMode, activeCards, col3Cards, fullRowCards, isDragging,
  onColReorder,   // <- no assertion in exposed-interface block
  onDragEnd, onFullRowDragEnd, onFullRowReorder, _fullRow,
  logoUrl, iconUrl, brandingMode, activeBrandingUrl, toggleBranding
})
```

**`EnhancedQuoteV3.spec.js` — `Exposed interface` describe block:**
```js
describe('Exposed interface', () => {
  test('test_EnhancedQuoteV3_exposes_required_properties', () => {
    // asserts: lastDataAt, isConnected, reconnecting, quoteData, manualTicker,
    // companyData, companyLoading, shortInterestData, shortInterestLoading,
    // activeCards, isDragging, logoUrl, iconUrl, brandingMode, activeBrandingUrl,
    // toggleBranding, fullRowCards, onFullRowDragEnd, onFullRowReorder, _fullRow
    // MISSING: onColReorder
  })
})
```

**Note:** `layoutMode`, `col3Cards`, and `onDragEnd` are asserted in other describe blocks (layout/settings tests) but also absent from the canonical exposed-interface block. Those are lower-priority since they have coverage elsewhere. `onColReorder` has no assertion anywhere in the spec.

## Why this matters

`onColReorder` is the external entry point for programmatic card reordering. If it disappears or changes signature, nothing fails in CI — the regression only surfaces at runtime in the dashboard. A contract assertion catches this at unit level.

## Recommended fix

Add to `test_EnhancedQuoteV3_exposes_required_properties`:

```js
expect(wrapper.vm.onColReorder).toBeDefined()
```

[...]

PR #174, opened 2026-04-14
feat(EQv3): per-card show/hide and normal/chips toggles

kuhl-haus-legion:
Implements the card toggles feature per design doc `EnhancedQuoteV3-Card-Toggles-Design.md`.

## What's in scope

All draggable cards except EQV3CompanyCard:
`today`, `prev`, `volume`, `session`, `short`

## Two toggles per card (edit mode only)

**👁 Show/Hide** — removes the card from the rendered layout. Hidden cards appear in a dashed tray below the columns so the user can re-show them without toggling lock/edit mode. Tray is invisible when locked.

**◧/▦ Normal/Chips** — switches between kv-list (default) and a compact chip layout. Applies in all layout modes (narrow, wide, full) — user preference, not layout-driven.

## Data model

Two new arrays in `settings` (backward-compatible; absent = all visible, all normal):
- `hiddenCards: string[]` — card IDs hidden by the user
- `chipCards: string[]` — card IDs in chip mode

## Architecture

- `hiddenCardIds` / `chipCardIds` — `Set<string>` computeds for O(1) lookup
- `visibleCards` — `activeCards` minus hidden; replaces `activeCards` in `col1Cards`, `col2Cards`, `fullRowCards`
- `hiddenCards` — cards that are hidden; drives tray
- `CARD_REGISTRY` — added `chipsCapable` flag; `company` is `false`

## Tests

13 new tests in a `Card visibility and chips toggles` describe block. 112/112 passing.

refs #175

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV3.vue
+++ b/client/src/components/widgets/EnhancedQuoteV3.vue
@@ -91,53 +91,102 @@
                 <div class="eqv3-card-label">
                   <span v-if="!isLocked" class="eqv3-drag-handle" title="Drag to reorder">⠿</span>
                   {{ card.label }}
+                  <span v-if="!isLocked" class="eqv3-card-controls">
+                    <button class="eqv3-card-toggle" :title="hiddenCardIds.has(card.id) ? 'Show card' : 'Hide card'" @click.stop="toggleCardVisibility(card.id)">{{ hiddenCardIds.has(card.id) ? '👁‍🗨' : '👁' }}</button>
+                    <button v-if="card.chipsCapable" class="eqv3-card-toggle" :title="chipCardIds.has(card.id) ? 'Normal mode' : 'Chips mode'" @click.stop="toggleCardChips(card.id)">{{ chipCardIds.has(card.id) ? '◧' : '▦' }}</button>
+                  </span>
                 </div>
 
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Implementation matches the design spec. All five in-scope cards have correct chip layouts. Session card reuses the existing `.eqv3-session-chips` layout as intended. Short interest handles loading/null in both chip and normal branches across all three template contexts (full-row, col1, col2).

`hiddenCardIds`/`chipCardIds` as Set computeds is the right choice for O(1) lookup. `visibleCards` correctly derived from `activeCards`, and all three column/row distribution computeds updated to use it. Toggle handlers follow the existing emit pattern. Card wrappers use dynamic `:class` binding so selector-based test assertions on `.eqv3-volume-card` etc. are valid.

All five new `defineExpose` entries (`visibleCards`, `hiddenCards`, `hiddenCardIds`, `chipCardIds`, `toggleCardVisibility`, `toggleCardChips`) are asserted in the exposed-interface block. Pre-existing gaps (`layoutMode`, `col3Cards`, `onColReorder`, `onDragEnd`) are unchanged — tracked in Issue #172.

13 new tests, correct emit capture pattern, observable behavior only. Approved.

oldschool-engineer (approved):

Issue #175, opened 2026-04-14
feat(EQv3): per-card show/hide and normal/chips toggles


kuhl-haus-legion:
## Summary

Each draggable card in EQv3 (except the Company card) gets two user-configurable toggles in edit mode:

1. **👁 Show/Hide** — hides the card from the layout. Hidden cards appear in a tray below the columns while in edit mode, so the user can re-show them without toggling lock on and off to see the result.

2. **◧/▦ Normal/Chips** — switches the card between its full kv-list rendering and a compact chip layout. User preference — applies in all layout modes (narrow, wide, full).

## In-scope cards

`Today`, `Previous Day`, `Volume`, `Session H/L`, `Short Interest`

## Out of scope

`Company` card — excluded from the chips toggle intentionally.

## Data model

Two new optional arrays in widget `settings` (fully backward-compatible; absent = all cards visible, all in normal mode):

```json
{
  "hiddenCards": ["short"],
  "chipCards": ["prev", "session"]
}
```

## Chip layouts per card

| Card | Chip fields |
|---|---|
| Today | O, VWAP |
| Previous Day | O, H, L, C, Vol, VWAP |
| Volume | Vol, Avg, Float, RVol |
| Session H/L | PRE H/L, REG H/L, AH H/L (existing chip layout) |
| Short Interest | SI, DTC, SVR |

PR #176, opened 2026-04-14
fix(EQv3): replace emoji toggles with filter-btn pills; fix RVol chip color

kuhl-haus-legion:
### Toggle buttons (fix #1)
Replaced emoji buttons (👁/👁‍🗨/◧/▦) with text-label pill buttons styled after NewsFeed's `.filter-btn` pattern:
- Default: dark bg (`#111`), `#333` border, `#888` text
- Hover: `#1a1a1a` bg, `#aaa` text
- **Active state** (hide=active, chips=active): `--pd-accent` border + text color so the current state is immediately readable against the dark background
- Labels: `hide`/`show` for visibility toggle, `list`/`chips` for render-mode toggle
- Tray restore button also updated (always renders in active style since it represents a hidden card)

### RVol chip color (fix #2)
Added `.eqv3-chip-val.extreme`, `.eqv3-chip-val.high`, `.eqv3-chip-val.medium` CSS rules mirroring the `.eqv3-rv-val` rules. The `relVolClass` computed was already applied to the chip span; the color just wasn't overriding `var(--text-primary)`.

## Tests
112/112 passing — no test changes needed (existing tests call methods directly, not emoji text).

refs #175

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV3.vue
+++ b/client/src/components/widgets/EnhancedQuoteV3.vue
@@ -92,8 +92,8 @@
                   <span v-if="!isLocked" class="eqv3-drag-handle" title="Drag to reorder">⠿</span>
                   {{ card.label }}
                   <span v-if="!isLocked" class="eqv3-card-controls">
-                    <button class="eqv3-card-toggle" :title="hiddenCardIds.has(card.id) ? 'Show card' : 'Hide card'" @click.stop="toggleCardVisibility(card.id)">{{ hiddenCardIds.has(card.id) ? '👁‍🗨' : '👁' }}</button>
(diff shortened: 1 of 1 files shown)


kuhl-haus-bishop (approved):
Both fixes are correct.

Toggle pills: three template sites updated consistently, tray restore button updated. Active state logic is right — `hide` button gets `--active` when the card is already hidden (i.e. action = show), `chips` button gets `--active` when chips mode is on. Non-blocking: the active state CSS uses `var(--pd-accent, #60a5fa)` and the hover uses `rgba(96,165,250,0.08)` — the fallback is blue while the actual token is violet (`#7c3aed`). In practice `base.css` always loads so this is never visible, but the fallback value is inconsistent. Worth fixing in a follow-up if it ever bothers anyone.

RVol chip color: three new rules mirroring `.eqv3-rv-val` exactly. The class was already applied, the specificity just was not there to override `var(--text-primary)`. Fix is correct.

Approved.

oldschool-engineer (approved):

PR #179, opened 2026-04-14
fix(EQv3): cap company card width to 50% in wide mode

kuhl-haus-legion:
Company card in wide mode was stretching to full column width (~half the widget), making it wider than the hero. Added `max-width: 50%` to `.eqv3-company-card` inside the `@container (min-width: 480px)` block.

112/112 tests passing.

refs #175


kuhl-haus-bishop (approved):
Correct placement and correct value. `max-width: 50%` inside the `480px` container block caps the company card to half the column width (quarter of the widget), consistent with the stated intent. No conflicts with the full-mode `min-width: 200px` rule — separate blocks. Approved.

oldschool-engineer (approved):

PR #180, opened 2026-04-14
Revert "fix(EQv3): cap company card width to 50% in wide mode (#179)"

kuhl-haus-legion:
Reverts #179. The `max-width: 50%` constraint was placed in the wide-mode container query, which broke narrow/wide layouts. Reverting cleanly until a targeted full-mode-only fix is designed.

refs #175


kuhl-haus-legion:
Closing — fixing forward instead.

PR #181, opened 2026-04-14
fix(EQv3): cap company card width to 50% in full mode only

kuhl-haus-legion:
Moves the `max-width: 50%` constraint from the wide-mode (480px) container query into the full-mode (960px) query, alongside the existing `min-width: 200px` rule on `.eqv3-full-row-draggable .eqv3-company-card`.

Narrow and wide layouts are unaffected.

112/112 tests passing.

refs #175


kuhl-haus-bishop (approved):
Correct. Wide-mode constraint removed; `max-width: 50%` added to the full-mode rule alongside the existing `min-width: 200px`. Both bounds now coexist on the same selector in the right block. Approved.

oldschool-engineer (approved):

PR #182, opened 2026-04-14
fix(EQv3): reduce hero width from 360px to 200px in full mode

kuhl-haus-legion:
Sets `.eqv3-hero` `width` to `200px` (down from `360px`) in the full-mode container query.

112/112 tests passing.

refs #175


kuhl-haus-legion:
Superseded by Tom's direct push.

Issue #183, opened 2026-04-15
feat(EQv3): three-column medium layout and sticky per-column card order


kuhl-haus-legion:
## Summary

Add a `medium` layout mode (720–1023px) with three draggable columns, and replace the current auto-split card distribution with sticky per-column card assignment that persists in settings.

Design doc: `Projects/EnhancedQuoteV3-Three-Column-Design.md` in Obsidian vault (Tom-approved, Bishop-approved).

## Breakpoints

| Mode | Width | Layout |
|---|---|---|
| `narrow` | < 360px | Single column, hero on top |
| `wide` | 360–719px | Two columns, hero on top |
| `medium` | 720–1023px | **Three columns**, hero on top ← NEW |
| `full` | ≥ 1024px | Hero left, horizontal card row right |

## Settings Shape Change

`cardOrder` (flat array) is replaced by `columns` (nested array):

```json
{
  "columns": [
    ["today", "prev"],
    ["volume", "session"],
    ["short", "company"]
  ]
}
```

A one-time migration shim converts old `cardOrder` saves on first load.

## Decisions

- Unassigned cards (not in any column) → silently append to col1
- Narrow fallback flattens all columns into col1 so no cards disappear
- `cardOrder` is internal-only; migration shim is sufficient
- Full-mode reorder collapses to col1 (simple/destructive — user controls auto-save)

## Current State

Implementation attempted in PRs #184, #185, #186 — all reverted via PR #187. Cross-column drag was broken in production despite passing tests. Root cause: vuedraggable cross-column drag event ordering (`@end` fires before `@update:model-value`) behaves differently in the real browser than in jsdom. Needs a revised approach before re-implementation.

## Scope

- `EnhancedQuoteV3.vue`: JS logic + CSS
- `EnhancedQuoteV3.spec.js`: ~15 new test cases
- Possibly `eqv3Utils.js` for migration shim extraction

kuhl-haus-legion:
Superseded by #188 (EQv4: grid-based widget with user-defined card layout). The three-column sticky-column approach is abandoned in favor of a fully user-defined inner grid.

PR #184, opened 2026-04-15
feat(EQv3): three-column medium layout and sticky per-column card order

kuhl-haus-legion:
Implements the design approved in `Projects/EnhancedQuoteV3-Three-Column-Design.md` (Tom-approved, Bishop-reviewed).

### New breakpoint
`MEDIUM: 720` added to `BREAKPOINTS`. Layout modes: narrow (<360px) → wide (360–719px) → **medium (720–1023px)** → full (≥1024px).

### Settings shape migration
`cardOrder` (flat array) is replaced by `columns` (nested array):
```json
{ "columns": [["today", "prev"], ["volume", "session"], ["short", "company"]] }
```
A `watch({ immediate: true })` migration shim converts legacy saves on first load. Idempotent — no-ops if `columns` is already present.

### Sticky per-column card assignment
- `col1Cards` / `col2Cards` / `col3Cards` read from `settings.columns[n]`
- Unassigned registry cards silently append to col1
- Cross-column drag updates all three columns in `onDragEnd`
- Full-mode reorder collapses all cards to col1 (intentional — user controls auto-save)

### Narrow fallback
`col1Cards` flattens all three columns at narrow width so no cards disappear when the widget is squeezed.

### col3 template
Three-column draggable block (identical card rendering to col1/col2), shown only when `layoutMode === 'medium'`.

### `defineExpose`
`col1Cards` and `col2Cards` added (previously internal-only).

## Tests
120/120 passing. 8 new tests:
- Migration shim (flat → columns, idempotent guard)
- col3Cards (medium mode, non-medium empty)
- col1Cards narrow flatten
- Cross-col drag col1→col3 with col2 preserved (Bishop #1)
- Hidden card in col3 absent from narrow flatten (Bishop #5)
- Unassigned card falls to col1
- Exposed API assertions for col1Cards / col2Cards / col3Cards

refs #183

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV3.vue
+++ b/client/src/components/widgets/EnhancedQuoteV3.vue
@@ -388,7 +388,7 @@
                           </draggable>
           </div>
 
-          <!-- Col 2: second half at wide -->
+          <!-- Col 2: visible at wide and medium -->
           <div v-if="!isNarrow" class="eqv3-col eqv3-col-2">
             <draggable
               :model-value="col2Cards"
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Both blocking items addressed correctly.

**Blocking #1 (cross-col drag col1→col3, col2 preserved):** Test is explicit — `_col2` is deliberately left null, triggering the `col2Cards` fallback, and the assertion confirms `columns[1]` is unchanged. This is exactly the right way to exercise the double-nextTick pattern for three columns.

**Blocking #2 (migration shim timing):** `watch(() => props.settings, shim, { immediate: true })` with an early return on `Array.isArray(settings.columns)`. Fires on mount, idempotent, drops `cardOrder` cleanly via `delete` on a spread copy (no prop mutation).

**Other items:** `BREAKPOINTS` constant updated to include `MEDIUM: 720` with matching CSS comment. Bishop #5 (hidden card in col3 absent from narrow flatten) has an explicit test. Unassigned card fallback to col1 tested. Narrow flatten deduplicates via `seen` Set — correct.

`col1Cards` and `col2Cards` added to `defineExpose` and asserted in the exposed-interface block. Pre-existing gaps (`layoutMode`, `onColReorder`, `onDragEnd`) unchanged — tracked in Issue #172.

Approved.

oldschool-engineer (approved):

PR #185, opened 2026-04-15
fix(EQv3): add min-height to columns so empty col3 has a drop target

kuhl-haus-legion:
vuedraggable cannot accept drops into a zero-height empty list — there's no DOM to drop onto. Adding `min-height: 48px` to `.eqv3-col` ensures all columns always have a visible drop zone, including col3 when it starts empty.

120/120 tests passing.

refs #183


kuhl-haus-bishop (approved):
Correct fix. `min-height: 48px` on `.eqv3-col` gives vuedraggable a drop target when col3 is empty. Applied to the column wrapper so it works for all columns, not just col3. Approved.

oldschool-engineer (approved):

PR #186, opened 2026-04-15
fix(EQv3): fix cross-column drag by emitting from onColReorder not onDragEnd

kuhl-haus-legion:
## Root cause

vuedraggable fires `@end` **before** `@update:model-value`. The previous implementation relied on `@end` (`onDragEnd`) to emit the updated column state, but `_col1/2/3` were always `null` at that point because `onColReorder` (`@update:model-value`) hadn't fired yet. Cross-column drags always fell back to stale `props.settings.columns`, so the second drag saw the same pre-drag state and appeared stuck.

## Fix

Move the emit into `onColReorder` (`@update:model-value`) — the same pattern already used by `onFullRowReorder` in full mode. For cross-column drags, vuedraggable fires `@update:model-value` twice (once per affected column); a `nextTick` debounce coalesces both fires into a single emit with the correct final state of all three columns.

`onDragEnd` is now a single-line `isDragging = false` setter.

## Tests

120/120 passing — no test changes needed; existing tests call `onColReorder` directly and await `nextTick`, which aligns with the new async behavior.

refs #183

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV3.vue
+++ b/client/src/components/widgets/EnhancedQuoteV3.vue
@@ -887,17 +887,20 @@ const _col2 = ref(null)
 const _col3 = ref(null)
 const _fullRow = ref(null)
 
-const onColReorder = (newVal, colNum) => {
+// onColReorder is called by @update:model-value, which fires AFTER @end per vuedraggable's
+// event order. For cross-column drags it fires once per affected column. We accumulate
+// the new column state here and schedule a single coalesced emit via nextTick.
+let _emitScheduled = false
+const onColReorder = async (newVal, colNum) => {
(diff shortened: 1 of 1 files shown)


kuhl-haus-bishop (approved):
Root cause is correct and fix is the right pattern — mirrors `onFullRowReorder` exactly. The `_emitScheduled` debounce coalesces the two `@update:model-value` fires from a cross-column drag into a single emit with the correct final state. `_col1/2/3` reset to null after the emit, `_emitScheduled` reset before it. `onDragEnd` reduced to `isDragging = false` — sets visual state only, emit path is fully owned by `onColReorder`. Approved.

oldschool-engineer (approved):

Issue #189, opened 2026-04-15
feat(EQv4): card SFCs — Hero, Today, PrevDay, Volume, Session, Short, Company


kuhl-haus-legion:
## Parent
refs #188

## Summary
Create 7 standalone card SFC files that the EQv4 root widget will render as inner `GridItem` children. Content is ported from EQv3 card sections + existing EQv3 card components. Cards are passive — they receive props only and know nothing about the grid, settings, or other cards.

## Files Created
- `EQV4HeroCard.vue` — symbol, price, change badge, branding (logo/icon)
- `EQV4TodayCard.vue` — open, VWAP, relative volume bar
- `EQV4PrevCard.vue` — prev day OHLCV + VWAP
- `EQV4VolumeCard.vue` — volume, avg vol, float, RVol
- `EQV4SessionCard.vue` — pre/reg/AH H/L
- `EQV4ShortCard.vue` — short interest, DTC, SVR
- `EQV4CompanyCard.vue` — description, sector, market cap, etc.

## Props Interface (per card)
Each card receives:
- `quoteData` — reactive quote object from WDS
- `isLocked` — Boolean (edit mode toggle)
- `chipsMode` — Boolean (chip/kv-list render mode; not all cards support it)
- Card-specific props as needed (e.g. `companyData`, `shortInterestData` for those cards)

Cards **never** read `settings` directly.

## Shared Utilities
Cards may reuse `eqv3Utils.js` (`fmtVol`, `fmt`) — no duplication.

## Test Cases (~7)
One spec per card: renders in normal mode, renders in chips mode where applicable, does not crash on null/empty `quoteData`.

## Size: L
7 files created, mostly porting existing EQv3 logic. No net-new data logic.

Issue #190, opened 2026-04-15
feat(EQv4): EnhancedQuoteV4.vue root widget


kuhl-haus-legion:
## Parent
refs #188

## Depends On
refs #189 (card SFCs must exist first)

## Summary
Implement the `EnhancedQuoteV4.vue` root widget: inner `GridLayout`, `settings` shape handling, default layout emission, `@layout-updated` → `update-settings`, WDS/company/short data fetching, widget bus connection, `brandingMode`, and `defineExpose`.

## Files Created
- `EnhancedQuoteV4.vue`
- `__tests__/EnhancedQuoteV4.spec.js`

## Key Logic
- On mount: if `settings.cards` absent/empty → render default layout and emit `update-settings` immediately
- `gridLayout` computed: maps `settings.cards` → layout array with `i: card.id`
- `@layout-updated` handler: strips `i`, emits `update-settings` with new card positions (fires on dragend/resizeend only — not on every drag tick)
- WDS via `useWebSocketClient` (`autoConnect: true`), company + short via REST
- Bus connection via `useWidgetBus` — Hero card is sole caller of `setActiveTicker`
- `chipCardIds`: computed Set from `settings.chipCards`

## Settings Shape
```json
{
  "gridCols": 6,
  "gridRowHeight": 40,
  "minOuterH": 12,
  "cards": [...],
  "chipCards": [],
  "brandingMode": "logo"
}
```

## defineExpose
`lastDataAt, isConnected, reconnecting, quoteData, manualTicker, companyData, companyLoading, shortInterestData, shortInterestLoading, gridLayout, gridCols, gridRowHeight, addCard, removeCard, activeCardIds`

## Test Cases (~12)
- Default layout emitted on first render when `settings.cards` absent
- `@layout-updated` emits `update-settings` with new card positions
- `@layout-updated` does not include `i` field in persisted cards
- WDS reconnect on ticker change
- Company data fetched on ticker change
- Short interest data fetched on ticker change
- `addCard('today')` places card at first available position
- `removeCard('hero')` removes card from `settings.cards`
- `gridCols` change emits `update-settings`
- `gridRowHeight` change emits `update-settings`
- `chipCardIds` computed correctly from `settings.chipCards`
- Exposed API assertions (all refs present)

## Size: L

PR #193, opened 2026-04-15
feat(EQv4): add 7 card SFCs with spec

kuhl-haus-legion:
## Summary

7 standalone card SFC files for EQv4. Each card is a passive props-only consumer — no knowledge of the grid, settings, or other cards. Content ported from EQv3.

refs #189

## Files Created
- `EQV4HeroCard.vue` — symbol, price, change badge, since-open, company identity, branding toggle (edit mode)
- `EQV4TodayCard.vue` — open, VWAP (list + chips)
- `EQV4PrevCard.vue` — prev day OHLCV + VWAP (list + chips)
- `EQV4VolumeCard.vue` — volume, avg vol, float, RVol bar + class (list + chips)
- `EQV4SessionCard.vue` — pre/reg/AH H/L (list + chips)
- `EQV4ShortCard.vue` — SI, DTC, SVR with loading/unavailable states (list + chips)
- `EQV4CompanyCard.vue` — kv rows + description expand/collapse, loading/unavailable states
- `__tests__/EnhancedQuoteV4Cards.spec.js` — 31 test cases, all passing

## Props Interface
All cards:
- `isLocked: Boolean` — edit mode toggle
- `chipsMode: Boolean` — chip/kv-list render mode (where applicable)
- Card-specific data props (`quoteData`, `shortInterestData`, `companyData`)

Cards **never** read `settings` directly.

## Shared Utilities
`fmtVol`, `truncateUrl`, `truncateDesc` imported from `eqv3Utils.js` — no duplication.

## Test Results
143/143 passing (112 existing + 31 new)

### DIFF
--- a/client/src/components/widgets/EQV4CompanyCard.vue
+++ b/client/src/components/widgets/EQV4CompanyCard.vue
@@ -1,0 +1,63 @@
+<template>
+  <div class="eqv4-card-body">
+    <div v-if="loading" class="eqv4-muted-msg">Company data loading...</div>
+    <div v-else-if="allNull" class="eqv4-muted-msg">Company data unavailable</div>
+    <div v-else>
+      <div class="eqv4-kv-list">
+        <div v-if="data.homepage_url" class="eqv4-kv">
+          <span class="eqv4-k">Web</span>
+          <a :href="data.homepage_url" target="_blank" rel="noopener noreferrer" class="eqv4-link">{{ truncateUrl(data.homepage_url) }}</a>
+        </div>
+        <div class="eqv4-kv"><span class="eqv4-k">Exchange</span><span class="eqv4-v">{{ data.primary_exchange || '—' }}</span></div>
+        <div class="eqv4-kv"><span class="eqv4-k">Mkt Cap</span><span class="eqv4-v">{{ data.market_cap != null ? '$' + fmtVol(data.market_cap) : '—' }}</span></div>
+        <div class="eqv4-kv"><span class="eqv4-k">Employees</span><span class="eqv4-v">{{ data.total_employees != null ? fmtVol(data.total_employees) : '—' }}</span></div>
+        <div class="eqv4-kv"><span class="eqv4-k">Listed</span><span class="eqv4-v">{{ data.list_date || '—' }}</span></div>
+      </div>
+      <div v-if="data.description" class="eqv4-company-desc-wrap">
+        <span class="eqv4-company-desc-text">
+          {{ expanded ? data.description : truncateDesc(data.description) }}
+        </span>
+        <span v-if="!expanded && truncateDesc(data.description) !== data.description">
+          <span class="eqv4-company-desc-ellipsis">… </span>
+          <button class="eqv4-see-more" @click="expanded = true">see more</button>
+        </span>
+        <button v-if="expanded" class="eqv4-see-more" @click="expanded = false"> less</button>
+      </div>
+    </div>
+  </div>
+</template>
+
+<script setup>
+import { ref, computed } from 'vue'
+import { truncateUrl, truncateDesc, fmtVol } from './eqv3Utils.js'
+
+const props = defineProps({
+  data:     { type: Object,  default: () => ({}) },
+  loading:  { type: Boolean, default: false },
+  isLocked: { type: Boolean, default: true },
+})
+
+const expanded = ref(false)
+
+const allNull = computed(() => {
+  const d = props.data
+  if (!d) return true
+  return d.name == null && d.sic_description == null && d.description == null && d.homepage_url == null
+})
+</script>
+
+<style scoped>
+.eqv4-card-body { height: 100%; overflow: auto; padding: 6px; box-sizing: border-box; }
+.eqv4-kv-list { display: flex; flex-direction: column; gap: 2px; }
+.eqv4-kv { display: flex; justify-content: space-between; align-items: center; padding: 2px 0; border-bottom: 1px solid var(--border, #2d2d3d); gap: 8px; }
+.eqv4-k { color: var(--text-muted, #64748b); font-size: 13px; font-family: system-ui, sans-serif; flex-shrink: 0; }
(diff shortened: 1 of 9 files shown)


kuhl-haus-bishop (changes requested):
## Review

3 blocking issues, 2 non-blocking. Do not merge until the blocking issues are resolved.

---

### 🔴 Blocking #1 — `EQV4CompanyCard`: prop named `data` instead of `companyData` — contract hole

`CompanyCard` declares `props: { data, loading, isLocked }`. Issue #189 spec says the card-specific prop is `companyData`. When the EQv4 root (PR #190) renders `<EQV4CompanyCard :company-data="companyData" />`, Vue maps `:company-data` → prop `companyData` — which doesn't exist. The card silently receives nothing and renders with an empty `data` object (the prop default `() => ({})`).

This will be invisible during development if you only test the card in isolation (as the current tests do). It will silently show "Company data unavailable" in production for every ticker.

**Fix:** Rename the prop to `companyData`. Update the spec accordingly.

---

### 🔴 Blocking #2 — `EQV4CompanyCard.allNull`: computed doesn't cover all rendered fields

`allNull` checks four fields: `name`, `sic_description`, `description`, `homepage_url`. The kv-list renders five different fields: `homepage_url`, `primary_exchange`, `market_cap`, `total_employees`, `list_date`.

`primary_exchange`, `market_cap`, `total_employees`, and `list_date` are not in the `allNull` guard. A ticker that has exchange + market cap + employees but no name/description will hit `allNull = true` and show "Company data unavailable" even though there's data to display.

**Fix:** `allNull` should check whether *any* of the rendered fields is non-null:

```js
const allNull = computed(() => {
  const d = props.data
  if (!d) return true
  return (
    d.homepage_url == null &&
    d.primary_exchange == null &&
    d.market_cap == null &&
    d.total_employees == null &&
    d.list_date == null &&
    d.name == null &&
    d.sic_description == null &&
    d.description == null
  )
})
```

Add a test: `with only primary_exchange set expect kv rows rendered (not unavailable message)`.

---

### 🔴 Blocking #3 — `EQV4HeroCard`: missing null/empty `quoteData` test

Issue #189 spec explicitly says each card must have a "does not crash on null/empty `quoteData`" test. `EQV4HeroCard` has no such test.

Additionally, `HeroCard` has inconsistent null handling internally: the `changeClass` computed guards against null quoteData (`if (!props.quoteData) return ''`), but the template accesses `quoteData.symbol`, `quoteData.close`, `quoteData.change`, `quoteData.pct_change` directly without optional chaining. If the root ever passes `null` (e.g. on unmount or ticker reset), the template throws a `TypeError` while the computed silently no-ops. Make the null handling consistent — either add optional chaining in the template for all direct field accesses, or remove the guard from `changeClass` and rely on the root's contract.

**Fix:**
1. Add optional chaining in template: `{{ quoteData?.symbol }}`, `{{ quoteData?.close }}`, etc.
2. Add test: `with empty quoteData object expect no crash and dashes rendered`.

---

### 🟡 Non-blocking #4 — `fmt()` duplicated across 6 files — export from `eqv3Utils.js`

`fmt()` is defined identically in `EQV4HeroCard`, `EQV4TodayCard`, `EQV4PrevCard`, `EQV4SessionCard`, `EQV4ShortCard`, and `EQV4VolumeCard` — six copies of:

```js
const fmt = (val, decimals = 2) => {
  const n = parseFloat(val)
  return isFinite(n) ? n.toFixed(decimals) : '—'
}
```

`eqv3Utils.js` already exports `fmtVol`, `truncateUrl`, `truncateDesc`. Add `fmt` there and import it. Six copies means any future change (e.g. locale formatting) requires touching 6 files.

---

### 🟡 Non-blocking #5 — `EQV4HeroCard`: `null` `end_timestamp` guard is untested

`dataAge` correctly returns `'—'` when `end_timestamp` is null/undefined. This guard is untested.

Add: `with null end_timestamp expect dataAge shows dash`.

---

## What's good

The overall structure is solid — cards are genuinely passive, props-only, no settings reads. Chip/list branching is clean throughout. VolumeCard's `floatShares` fallback (`free_float ?? share_class_shares_outstanding`) is correct. The `toggle-branding` event pattern using `attrs: { 'onToggle-branding': ... }` is right for VTU 2.4.x with `<script setup>`. CompanyCard's expand/collapse is well-tested. 31 tests at 143/143 passing is a good sign — just needs the three fixes above before merge.

kuhl-haus-bishop (commented):
## Updated Review — Clarification on #4 and #5

In my initial review I labeled issues #4 and #5 as "non-blocking." I'm correcting that now.

**Both #4 and #5 are blocking — fix them in this PR.**

Clarifying the routing rule going forward: blocking = fix in this PR; non-blocking = separate issue, scheduled for later. There is no middle ground.

---

### 🔴 Blocking #4 (revised) — `fmt()` duplicated across 6 files — fix in this PR

`fmt()` is defined identically in all six card SFCs. `eqv3Utils.js` is the right home — `fmtVol` is already there. Add `export const fmt = ...` to `eqv3Utils.js` and import it in all six cards. This is mechanical, trivially reviewable, and the right time to do it is before 6 more files get written that copy the same function.

---

### 🔴 Blocking #5 (revised) — `dataAge` null guard untested — fix in this PR

The guard exists and is correct. It has no test coverage. The issue spec says each card needs a "does not crash on null/empty `quoteData`" test — the missing timestamp case is the real pre-connect scenario for `dataAge`. Add: `with null end_timestamp expect dataAge shows dash`.

kuhl-haus-bishop (approved):
All five issues resolved. Approved.

- `EQV4CompanyCard` prop renamed to `companyData` — contract hole closed
- `allNull` now covers all 8 rendered fields; new test for `primary_exchange`-only case directly covers the bug
- `EQV4HeroCard` template uses `quoteData?.field` throughout; null handling is consistent with `changeClass`; empty `{}` test and null `end_timestamp` test both added
- `fmt()` exported from `eqv3Utils.js` with JSDoc, imported in all 6 card SFCs — no local copies remaining
- `eqv3Utils.js` diff is clean: `fmt` inserted before `fmtVol`, consistent style with existing exports

Good execution on the turnaround. Ready for #190.

oldschool-engineer (approved):

Issue #191, opened 2026-04-15
feat(EQv4): edit mode — EQV4CardPicker, add/remove card, grid config panel


kuhl-haus-legion:
## Parent
refs #188

## Depends On
refs #190 (root widget must exist first)

## Summary
Implement edit mode UX for EQv4: the card picker, X-button removal, and the grid config panel (columns + row height). All edit mode controls are visible only when `!isLocked`.

## Files Created
- `EQV4CardPicker.vue` — dropdown/modal listing cards not currently on the grid; selecting a card calls `addCard(cardId)`

## Files Modified
- `EnhancedQuoteV4.vue` — wire in picker, X buttons on card headers, grid config inputs

## Key Logic
- **X button** on each card header: calls `removeCard(card.id)`, emits `update-settings`
- **Add card button**: opens `EQV4CardPicker`; picker lists only cards absent from `settings.cards`
- **First-fit scan** (row-major): scan left-to-right, top-to-bottom; place card at first `(x, y)` that fits within `gridCols` without overlap; fallback: `x=0, y=maxOccupiedY+1`
- **Grid config panel**: numeric input for columns (min 1, max 24) and row height (min 20, max 120); each change emits `update-settings`

## Test Cases (~6)
- Picker lists only absent cards (not already-placed cards)
- Selecting a card from picker appends it at first-fit position
- First-fit fallback: append to bottom when no gap fits default `w`
- X button removes card and emits `update-settings`
- Column input change emits `update-settings` with new `gridCols`
- Row height input change emits `update-settings` with new `gridRowHeight`

## Size: M

kuhl-haus-legion:
All deliverables from this issue were implemented as part of PR #194 (refs #190):

- **EQV4CardPicker.vue** — created; dropdown listing absent cards, emits `add-card`, closes on selection
- **X button on card headers** — wired in `EnhancedQuoteV4.vue`, calls `removeCard(card.id)`
- **First-fit scan** (row-major + fallback) — implemented in `addCard()`
- **Grid config panel** — columns (1–24) and row height (20–120) inputs, both emit `update-settings`
- All 6 specified test cases present and passing in `EnhancedQuoteV4.spec.js`

Closing as complete.

PR #194, opened 2026-04-15
feat(EQv4): add EnhancedQuoteV4 root widget and EQV4CardPicker

kuhl-haus-legion:
## Summary

Implements the `EnhancedQuoteV4.vue` root widget and `EQV4CardPicker.vue`.

refs #190

## Files Created
- `EnhancedQuoteV4.vue` — root widget: inner `GridLayout`, settings handling, data layer, edit mode wiring, defineExpose
- `EQV4CardPicker.vue` — add-card dropdown listing absent cards
- `__tests__/EnhancedQuoteV4.spec.js` — 22 test cases, all passing

## Key Logic

**Settings / layout:**
- `gridLayout` computed: maps `settings.cards` → layout array with `i: card.id`; default 7-card layout if absent
- First render without `settings.cards`: emits `update-settings` with default layout immediately (immediate watcher)
- `@layout-updated` (dragend/resizeend only): strips `i`, persists new card positions via `update-settings`
- Internal `internalLayout` ref synced from `gridLayout` computed on settings change

**Edit mode:**
- X button on card header calls `removeCard(cardId)`
- EQV4CardPicker lists absent cards; selection calls `addCard(cardId)`
- First-fit scan (row-major, append-to-bottom fallback) places new cards
- Cols/row-height numeric inputs emit `update-settings` on change (clamped: cols 1–24, rowHeight 20–120)

**Data layer:**
- WDS via `useWebSocketClient` (`autoConnect: false`, arm on config ready)
- Company + short interest via Massive REST API (same as EQv3)
- Bus connection via `useWidgetBus` — Hero card is sole `setActiveTicker` caller
- `brandingMode` + `activeBrandingUrl` + `toggleBranding` match EQv3 shape

**defineExpose:**
`lastDataAt, isConnected, reconnecting, quoteData, manualTicker, companyData, companyLoading, shortInterestData, shortInterestLoading, gridLayout, gridCols, gridRowHeight, chipCardIds, addCard, removeCard, activeCardIds`

## Test Results
168/168 passing (146 existing + 22 new)

### DIFF
--- a/client/src/components/widgets/EQV4CardPicker.vue
+++ b/client/src/components/widgets/EQV4CardPicker.vue
@@ -1,0 +1,77 @@
+<template>
+  <div class="eqv4-picker">
+    <button class="eqv4-picker-btn filter-btn" @click="open = !open">+ Add card</button>
+    <div v-if="open" class="eqv4-picker-dropdown">
+      <button
+        v-for="card in absentCards"
+        :key="card.id"
+        class="eqv4-picker-item"
+        @click="select(card.id)"
+      >{{ card.label }}</button>
+    </div>
+  </div>
+</template>
+
+<script setup>
+import { ref } from 'vue'
+
+defineProps({
+  absentCards: { type: Array, required: true },
+})
+
+const emit = defineEmits(['add-card'])
+const open = ref(false)
+
+const select = (cardId) => {
+  open.value = false
+  emit('add-card', cardId)
+}
+</script>
+
+<style scoped>
+.eqv4-picker { position: relative; }
+.filter-btn {
+  background: none;
+  border: 1px solid var(--border, #2d2d3d);
+  border-radius: 4px;
+  color: var(--text-muted, #64748b);
+  font-size: 11px;
+  padding: 3px 8px;
+  cursor: pointer;
+  font-family: system-ui, sans-serif;
+  transition: border-color 0.15s, color 0.15s;
+}
+.filter-btn:hover {
+  border-color: var(--pd-accent, #7c3aed);
+  color: var(--pd-accent, #7c3aed);
+}
+.eqv4-picker-dropdown {
+  position: absolute;
+  bottom: calc(100% + 4px);
+  left: 0;
+  background: var(--surface, #141420);
+  border: 1px solid var(--border, #2d2d3d);
+  border-radius: 4px;
+  min-width: 140px;
+  z-index: 100;
+  display: flex;
+  flex-direction: column;
+  gap: 1px;
+  padding: 4px;
+}
+.eqv4-picker-item {
+  background: none;
+  border: none;
+  color: var(--text-primary, #e2e8f0);
+  font-size: 12px;
+  padding: 5px 8px;
+  cursor: pointer;
+  text-align: left;
+  border-radius: 3px;
+  font-family: system-ui, sans-serif;
+}
+.eqv4-picker-item:hover {
+  background: var(--bg, #0d0d12);
+  color: var(--pd-accent, #7c3aed);
+}
+</style>
--- a/client/src/components/widgets/EnhancedQuoteV4.vue
+++ b/client/src/components/widgets/EnhancedQuoteV4.vue
@@ -1,0 +1,632 @@
+<template>
+  <div class="eqv4-widget" ref="widgetEl">
+    <!-- Ticker input bar — always visible -->
+    <div class="eqv4-controls">
+      <input
+        v-model="inputTicker"
+        class="eqv4-input"
+        placeholder="Enter ticker (e.g. AAPL)"
+        maxlength="10"
+        spellcheck="false"
+        @keyup.enter="applyInput"
+        @keyup.escape="inputTicker = ''"
+      />
+      <button class="eqv4-go-btn" @click="applyInput" @touchend.prevent="applyInput" title="Load quote">Go</button>
+    </div>
+
+    <!-- No ticker yet -->
+    <div v-if="!activeTicker" class="eqv4-empty">
+      <span class="eqv4-empty-icon">⚡</span>
+      <span class="eqv4-empty-text">Enter a ticker above, or link to a scanner and click a row</span>
+    </div>
+
+    <!-- Ticker set, waiting for data -->
+    <div v-else-if="!quoteData" class="eqv4-empty">
+      <span class="eqv4-empty-icon">⏳</span>
+      <span class="eqv4-empty-text">Waiting for data for <strong>{{ activeTicker }}</strong>...</span>
+    </div>
+
+    <!-- Quote data available -->
+    <div v-else class="eqv4-body">
+      <GridLayout
+        v-model:layout="internalLayout"
+        :col-num="gridCols"
+        :row-height="gridRowHeight"
+        :margin="[5, 5]"
+        :is-draggable="!isLocked"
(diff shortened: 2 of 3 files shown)


kuhl-haus-bishop (changes requested):
## Review

3 blocking issues. Do not merge until all three are resolved.

The implementation is solid — default layout emission, `i`/`id` mapping, first-fit scan, deferred WDS connect, ticker change flow are all correct. The gaps are in test coverage.

---

### 🔴 Blocking #1 — `@layout-updated` handler (`onLayoutUpdated`) is untested

`onLayoutUpdated` is the most critical function in this PR — it's the persistence mechanism for all grid state. Issue #190's spec explicitly requires two tests for it:

> - `@layout-updated` emits `update-settings` with new card positions
> - `@layout-updated` does not include `i` field in persisted cards

Neither test exists. The `i`-stripping logic (`newLayout.map(({ i, ...rest }) => ({ ...rest, id: i }))`) is correct, but it's untested, which means a future refactor could silently break persistence.

**Fix:** trigger the layout-updated event via the mock GridLayout component and assert on the emitted payload:

```js
test('with layout-updated event expect update-settings emitted without i field', async () => {
  // Arrange
  const emitted = []
  const wrapper = mountWidget(
    { settings: { cards: [{ id: 'hero', x: 0, y: 0, w: 6, h: 3 }] } },
    (s) => emitted.push(s)
  )
  wrapper.vm.quoteData = SAMPLE_QUOTE
  await nextTick()

  // Act
  const grid = wrapper.findComponent({ name: 'GridLayout' })
  await grid.vm.$emit('layout-updated', [{ i: 'hero', x: 1, y: 0, w: 5, h: 3 }])
  await nextTick()

  // Assert
  const payload = emitted[emitted.length - 1]
  expect(payload.cards[0].id).toBe('hero')
  expect(payload.cards[0].x).toBe(1)
  expect(payload.cards[0].i).toBeUndefined()
})
```

---

### 🔴 Blocking #2 — `EQV4CardPicker` open/close behavior untested

`EQV4CardPicker` is created in this PR. Its core behaviors — opening the dropdown on button click and closing it after a card is selected — are untested. `select()` sets `open.value = false` and emits `add-card`. If either step broke, no test would catch it.

**Fix:** add a `describe('EQV4CardPicker')` block in the spec (can be a direct unit test of the component):

```js
import EQV4CardPicker from '../EQV4CardPicker.vue'

describe('EQV4CardPicker', () => {
  const absentCards = [{ id: 'today', label: 'Today' }, { id: 'volume', label: 'Volume' }]

  test('with add-card button click expect dropdown opens', async () => {
    const wrapper = mount(EQV4CardPicker, { props: { absentCards } })
    expect(wrapper.find('.eqv4-picker-dropdown').exists()).toBe(false)
    await wrapper.find('.eqv4-picker-btn').trigger('click')
    expect(wrapper.find('.eqv4-picker-dropdown').exists()).toBe(true)
  })

  test('with card item click expect add-card emitted and dropdown closes', async () => {
    const calls = []
    const wrapper = mount(EQV4CardPicker, {
      props: { absentCards },
      attrs: { 'onAdd-card': (id) => calls.push(id) },
    })
    await wrapper.find('.eqv4-picker-btn').trigger('click')
    await wrapper.find('.eqv4-picker-item').trigger('click')
    expect(calls).toEqual(['today'])
    expect(wrapper.find('.eqv4-picker-dropdown').exists()).toBe(false)
  })
})
```

---

### 🔴 Blocking #3 — `addCard` fallback test is a false positive

Test: `'with addCard called and no gap fits expect card appended to bottom'`

This test does not exercise the fallback. Here's why:

- Setup: hero at `x=0, y=0, w=6, h=3` in a 6-column grid. Adding `'today'` (`defaultW: 2`).
- Row `y=3` is entirely unoccupied. `today` fits there via first-fit at `(x=0, y=3)`.
- The test asserts `added.y >= 3` — which passes — but it's testing first-fit, not the fallback.

The fallback code path (`placed = { id: cardId, x: 0, y: maxY, ... }`) is never reached. The test passes but the fallback is untested.

**Fix:** construct a scenario where no row has a gap wide enough for the new card's `defaultW`. For example: `gridCols: 3` with a card at `x: 0, y: 0, w: 3, h: 2`, then add `'session'` (`defaultW: 3`) — the only fit is `y=0..1` which is fully occupied, so fallback to `y=2`.

Or more explicitly: fill multiple rows completely so no gap exists anywhere, then assert `added.x === 0` and `added.y === maxOccupiedY`.

kuhl-haus-bishop (changes requested):
## Re-review

Two of three blocking issues fully resolved. One remains.

**✅ #1 — `onLayoutUpdated` tests:** Both tests are present and correct. `findComponent({ name: 'GridLayout' })` works against the named mock, `$emit('layout-updated', [...])` triggers the handler, and the two assertions (new positions emitted, `i` field absent from persisted cards) are exactly right.

**✅ #2 — `EQV4CardPicker` tests:** Three tests covering open, emit+close, and label rendering. The `attrs: { 'onAdd-card': ... }` pattern is correct for VTU 2.4.x + `<script setup>`. Clean.

---

### 🔴 Blocking #3 (still) — `addCard` fallback test is still a false positive

The updated test uses `gridCols: 3`, `hero` at `x: 0, y: 0, w: 3, h: 2`, adding `'session'` (`defaultW: 3`).

Here's what actually happens:
- `hero` occupies all cells in `y=0` and `y=1` (full 3 columns, 2 rows)
- `maxY = hero.y + hero.h = 2`
- The scan runs `y = 0..2`
- At `y=2`: `session` has `defaultH=3`, so it checks `dy=0..2` at `x=0` — none of those cells (`y=2`, `y=3`, `y=4`) are occupied → **fits at first-fit `(x=0, y=2)`**
- `placed` is set by the first-fit scan, **not** by the fallback

The fallback (`placed = { id: cardId, x: 0, y: maxY, ... }`) only fires when `placed === null` after the entire scan. The scan finds a fit at `y=2`, so `placed` is never `null`. The fallback code is still untested.

**One-line fix:** change `gridCols: 3` to `gridCols: 2` and `hero` width to `w: 2`. Now when you add `'session'` (`defaultW: 3`), the inner loop condition `x <= cols - def.defaultW` evaluates to `x <= 2 - 3 = -1` — the loop body never runs, `placed` stays `null`, and the fallback fires. Observable result is identical (`x: 0, y: 2`) but the fallback path is actually exercised.

```js
// Change this:
{ settings: { gridCols: 3, cards: [{ id: 'hero', x: 0, y: 0, w: 3, h: 2 }] } }
// To this:
{ settings: { gridCols: 2, cards: [{ id: 'hero', x: 0, y: 0, w: 2, h: 2 }] } }
```

The rest of the test (including the assertion `added.x === 0, added.y === 2`) stays exactly as written. One line change.

kuhl-haus-bishop (approved):
Approved. All three blocking issues resolved.

The fallback test is now correct: `gridCols: 2`, `hero` at `w: 2`, adding `'session'` (`defaultW: 3`). Loop condition `x <= 2-3 = -1` never executes, `placed` stays `null`, fallback fires. The comment in the test explains exactly why — good. That's the right level of documentation for a non-obvious edge case.

Full picture: `onLayoutUpdated` coverage, CardPicker open/close/emit, and the first-fit vs fallback distinction are all properly exercised. Ready for #191.

oldschool-engineer (approved):

Issue #192, opened 2026-04-15
feat(EQv4): register widget in WidgetMenu and WidgetWrapper


kuhl-haus-legion:
## Parent
refs #188

## Depends On
refs #191 (EQv4 must be complete and passing before wiring into dashboard)

## Summary
Register EQv4 in the two files that expose it to the dashboard UI.

## Files Modified
- `WidgetMenu.vue` — add EQv4 to the widget picker list
- `WidgetWrapper.vue` — add EQv4 to the component resolver

## Test Cases (~2)
- Widget menu includes EQv4 as a selectable option
- WidgetWrapper resolves the EQv4 component by type key

## Size: XS

Issue #188, opened 2026-04-15
feat(EQv4): grid-based widget with user-defined card layout


kuhl-haus-legion:
## Summary

EQv4 is a new widget (coexisting with EQv3) that replaces hard-coded responsive breakpoints with a user-defined inner `vue3-grid-layout-next` grid. Cards are placed, sized, and arranged by the user in edit mode. No column-count constraints, no vuedraggable cross-list coordination.

Design doc: `Projects/EnhancedQuoteV4-Design.md` in Obsidian vault (Tom ✅ Bishop ✅).

## Architecture

- EQv4 occupies one outer `GridItem` and contains its own inner `GridLayout` instance
- 7 card SFCs: Hero, Today, Previous Day, Volume, Session H/L, Short Interest, Company
- Cards are props-only consumers — root widget owns all data fetching and bus connection

## Settings Shape

```json
{
  "gridCols": 6,
  "gridRowHeight": 40,
  "minOuterH": 12,
  "cards": [
    { "id": "hero", "x": 0, "y": 0, "w": 6, "h": 3 },
    ...
  ],
  "chipCards": [],
  "brandingMode": "logo"
}
```

`id` is the canonical card identifier; `i` is derived from `id` in a computed before passing to GridLayout.

## Edit Mode

- **X button** on each card header removes the card
- **Add card** picker lists absent cards; first-fit placement (row-major, append-to-bottom fallback)
- **Drag/resize** via `vue3-grid-layout-next` native — no vuedraggable
- **Grid config** panel: columns + row height (numeric inputs)
- `@layout-updated` drives `update-settings` emit (fires on dragend/resizeend only)

## Chip render mode

Cards that support chip/kv-list in EQv3 support it in EQv4. Root passes `chipsMode` prop; cards never read `settings` directly.

---

## Implementation Plan

Work is broken into 4 sub-issues in execution order. Each is independently reviewable.

| # | Sub-issue | Size | Depends on |
|---|-----------|------|------------|
| #189 | Card SFCs (7 files: Hero, Today, PrevDay, Volume, Session, Short, Company) | L | — |
| #190 | `EnhancedQuoteV4.vue` root widget (inner GridLayout, settings handling, data layer, defineExpose) | L | #189 |
| #191 | Edit mode — `EQV4CardPicker`, add/remove card, grid config panel | M | #190 |
[...]

PR #195, opened 2026-04-15
feat(EQv4): register enhanced-quote-v4 in WidgetMenu and WidgetWrapper

kuhl-haus-legion:
## Summary

Wires EQv4 into the dashboard UI.

refs #192
refs #188

## Files Modified
- `WidgetMenu.vue` — adds `enhanced-quote-v4` entry (label: "Enhanced Quote V4", icon: 💎)
- `WidgetWrapper.vue` — imports `EnhancedQuoteV4` and adds `'enhanced-quote-v4'` to the component resolver map

## Files Created
- `__tests__/WidgetMenuAndWrapper.spec.js` — 5 test cases covering WidgetMenu EQv4 listing + emit, and WidgetWrapper resolution for EQv4, EQv3, and the backward-compat `enhanced-quote` alias

## Test Results
178/178 passing (173 existing + 5 new)

### DIFF
--- a/client/src/components/WidgetMenu.vue
+++ b/client/src/components/WidgetMenu.vue
@@ -35,7 +35,7 @@ const availableWidgets = [
   { type: 'company-news', label: 'Company News', icon: '🗞️' },
   { type: 'quote', label: 'Quote', icon: '⚡' },
   { type: 'enhanced-quote', label: 'Enhanced Quote', icon: '✨' },
-  { type: 'enhanced-quote-v4', label: 'Enhanced Quote V4', icon: '⚡' },
+  { type: 'enhanced-quote-v4', label: 'Enhanced Quote V4', icon: '💎' },
 ]
 
 const selectWidget = (widgetType) => {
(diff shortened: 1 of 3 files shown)


kuhl-haus-bishop (approved):
Approved.

The wiring is correct: `WidgetMenu` entry, `WidgetWrapper` import + resolver key, and 5 tests covering the new type plus the existing aliases. Nothing to block on.

Two minor observations, no action required:

1. `enhanced-quote-v4` uses the same `⚡` icon as the plain `quote` widget. `enhanced-quote` (EQv3) uses `✨`. With two `⚡` entries in the menu, the icons won't help users distinguish them. Consider a distinct icon for EQv4 — `🔲` or `⚡v4` — but this is a UX call, not a correctness issue.

2. WidgetMenu test 1 has two dead-code lines (`buttons`/`types` are defined but never used; the assertion goes straight to `wrapper.text()`). Harmless, but they can be deleted.

Neither item is worth a follow-up issue. Close #192 and #188 after merge.

kuhl-haus-bishop (changes requested):
Retracting approval. Blocking issue I should have caught the first time.

### 🔴 Blocking — Dead code in `WidgetMenuAndWrapper.spec.js`

Test `'with menu open expect enhanced-quote-v4 option present'` contains two lines that are defined and immediately discarded:

```js
const buttons = wrapper.findAll('.widget-button')
const types = buttons.map(b => b.attributes('data-type') ?? b.text())
```

Neither `buttons` nor `types` is referenced again. The assertion goes straight to `wrapper.text()`. Delete both lines. The test is valid without them — they're just noise that shouldn't have been committed.

kuhl-haus-bishop (approved):
Clean. Approved.

oldschool-engineer (approved):

PR #196, opened 2026-04-15
fix(EQv4): move edit bar outside data guard; fix layout-update feedback loop

kuhl-haus-legion:
## Summary

Two production bugs found during first use of EQv4 after all sub-issues merged.

refs #188

## Bug 1 — Edit tools not accessible before a ticker is set

**Symptom:** No controls to add cards or configure the grid on a freshly added EQv4 widget until a ticker is entered *and* quote data arrives.

**Root cause:** `eqv4-edit-bar` was nested inside `v-else class="eqv4-body"`, gated on `activeTicker && quoteData`. In the empty and waiting states, the entire edit section was hidden.

**Fix:** Moved the edit bar outside the data guard — it now renders whenever `!isLocked`, regardless of ticker or data state.

## Bug 2 — UI lockup when linked to a live ticker on the color bus

**Symptom:** Linking EQv4 to a live ticker via the color bus locked up the entire UI.

**Root cause:** `watch(gridLayout, { deep: true })` reset `internalLayout` on every settings change. `onLayoutUpdated` emits `update-settings` → parent stores new settings → `props.settings` changes → `gridLayout` computed reacts → watcher fires → `internalLayout` reset → GridLayout fires `@layout-updated` again → infinite loop.

**Fix:** Added `_ownLayoutUpdate` guard flag. Set to `true` before emitting in `onLayoutUpdated`. The watcher checks and clears it — skips the reset when the change originated from us. External settings changes (parent prop updates) still sync through normally.

## Files Changed
- `EnhancedQuoteV4.vue` — edit bar moved, guard flag added
- `EnhancedQuoteV4.spec.js` — grid config tests updated to reflect edit bar is always visible when unlocked (no ticker/data setup required)

## Test Results
178/178 passing

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV4.vue
+++ b/client/src/components/widgets/EnhancedQuoteV4.vue
@@ -28,6 +28,9 @@
 
     <!-- Quote data available -->
     <div v-else class="eqv4-body">
+
+      <!-- Edit bar — always visible when unlocked, regardless of data state -->
+
       <GridLayout
         v-model:layout="internalLayout"
         :col-num="gridCols"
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
## Review

One blocking issue. Both fixes are mechanically correct — this is purely a test coverage gap for the bug that was just fixed.

---

### 🔴 Blocking — Missing regression test for the `_ownLayoutUpdate` feedback loop fix

Bug 2 was a production defect: `onLayoutUpdated` → `emit(update-settings)` → `props.settings` update → `gridLayout` computed → watcher fires → `internalLayout` reset → `@layout-updated` fires again → infinite loop. The `_ownLayoutUpdate` flag correctly breaks it. There is no test that guards against this regression.

The `onLayoutUpdated` tests added in PR #194 verify that the correct payload is emitted and that `i` is stripped. They do not verify that `update-settings` is emitted exactly once per drag event. Without that assertion, the feedback loop could silently return and no test would catch it.

**Fix:** Add to the `onLayoutUpdated` describe block:

```js
test('with layout-updated event expect update-settings emitted exactly once (no feedback loop)', async () => {
  // Arrange
  const emitted = []
  const wrapper = mountWidget(
    { settings: { cards: [{ id: 'hero', x: 0, y: 0, w: 6, h: 3 }] } },
    (s) => emitted.push(s)
  )
  await nextTick()
  wrapper.vm.quoteData = SAMPLE_QUOTE
  await nextTick()
  const beforeCount = emitted.length

  // Act — simulate one drag-end event
  const grid = wrapper.findComponent({ name: 'GridLayout' })
  await grid.vm.$emit('layout-updated', [{ i: 'hero', x: 1, y: 0, w: 5, h: 3 }])
  await nextTick()
  await nextTick() // allow any potential echo-back to propagate

  // Assert — exactly one emission per drag event
  expect(emitted.length - beforeCount).toBe(1)
})
```

---

### ✅ What's correct

**Bug 1:** Edit bar is now a sibling of the data guard blocks, not nested inside `eqv4-body`. Grid config and card picker are accessible in all states when `!isLocked`. Spec tests correctly simplified — no ticker/quoteData setup needed for the grid config tests.

**Bug 2:** `_ownLayoutUpdate` as a non-reactive `let` is the right choice — reactive ref would add unnecessary tracking overhead. Set synchronously before emit, cleared synchronously in the watcher callback. Single-threaded JS guarantees no race condition between the two. The logic is correct; it just needs a test.

kuhl-haus-bishop (approved):
Approved. Regression test is correct — two `nextTick()` waits give the echo-back room to propagate before asserting count. Both bugs fixed, both covered.

oldschool-engineer (approved):

PR #197, opened 2026-04-15
feat(EQv4): always render grid; hero wide/narrow toggle; chips toggle in card header

kuhl-haus-legion:
## Summary

Three UX improvements to EQv4 based on first-use feedback.

refs #188

## Fix 1 — Grid always rendered (no longer gated on ticker/data)

**Before:** Widget was empty until a ticker was set *and* quote data arrived. The grid and cards only rendered inside `v-else` (quoteData guard).

**After:** Grid renders immediately on mount. Empty/waiting states are now an overlay (`eqv4-overlay`) that appears *above* the grid using `position: absolute` with a semi-transparent background. Cards receive `quoteData=null` until data arrives — all card templates updated with optional chaining (`?.`) and `default: null` prop definitions to handle this safely.

## Fix 2 — Hero card wide/narrow layout toggle

`EQV4HeroCard` now supports a `heroMode` prop (`'wide' | 'narrow'`, default `'wide'`):

- **wide**: vertical stack — symbol block → price block → identity block (same as current)
- **narrow**: two-column — left column has logo/symbol/flame/company name/SIC; right column has price/change/since-open/as-of

Toggle button appears on the Hero card header in edit mode. State stored in `settings.heroMode`. `toggleHeroMode()` exposed on the widget.

## Fix 3 — Chips/kv-list toggle in card headers

Each card header in edit mode now shows a `chips`/`list` toggle button for `chipsCapable` cards (today, prev, volume, session, short). Active state uses `--pd-accent` border + text, matching EQv3 style. `toggleCardChips(cardId)` exposed on the widget.

## Files Changed
- `EnhancedQuoteV4.vue` — overlay pattern, card header controls, heroMode/chips toggle handlers, defineExpose additions
- `EQV4HeroCard.vue` — wide/narrow layout modes with CSS class switching
- `EQV4TodayCard.vue`, `EQV4PrevCard.vue`, `EQV4VolumeCard.vue`, `EQV4SessionCard.vue` — `quoteData` prop changed to `default: null`, optional chaining on all field accesses
- `EnhancedQuoteV4.spec.js` — updated overlay tests, 10 new test cases (heroMode, toggleCardChips, EQV4HeroCard layout modes)

## Test Results
189/189 passing

### DIFF
--- a/client/src/components/widgets/EQV4HeroCard.vue
+++ b/client/src/components/widgets/EQV4HeroCard.vue
@@ -1,50 +1,90 @@
 <template>
-  <div class="eqv4-hero-card">
-    <!-- Symbol + branding image -->
-    <div class="eqv4-hero-symbol-block">
-      <img
-        v-if="activeBrandingUrl"
-        :src="activeBrandingUrl"
-        class="eqv4-hero-logo"
-        :alt="brandingMode === 'icon' ? 'Company icon' : 'Company logo'"
-      />
-      <div class="eqv4-hero-symbol-row">
-        <span class="eqv4-symbol">{{ quoteData?.symbol }}</span>
-        <img v-if="flameIcon" :src="flameIcon.src" :title="flameIcon.tooltip" class="eqv4-flame-icon" />
-      </div>
-      <!-- Branding toggle — edit mode only -->
-      <button
-        v-if="!isLocked"
-        class="eqv4-branding-toggle filter-btn"
-        :title="brandingMode === 'logo' ? 'Switch to icon' : 'Switch to logo'"
-        @click="emit('toggle-branding')"
-      >{{ brandingMode === 'logo' ? 'icon' : 'logo' }}</button>
-    </div>
+  <div :class="['eqv4-hero-card', heroMode === 'wide' ? 'eqv4-hero--wide' : 'eqv4-hero--narrow']">
 
-    <!-- Price block -->
-    <div class="eqv4-hero-price-block">
-      <span class="eqv4-price">${{ fmt(quoteData?.close, 2) }}</span>
-      <span :class="['eqv4-change-badge', changeClass]">
-        {{ (quoteData?.change ?? 0) >= 0 ? '+' : '' }}{{ fmt(quoteData?.change, 2) }}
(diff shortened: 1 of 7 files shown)


kuhl-haus-bishop (changes requested):
## Review

One blocking issue. Everything else is clean.

---

### 🔴 Blocking — `:hero-mode` passed to all card components, not just HeroCard

```vue
<component
  :is="cardComponent(item.i)"
  ...
  :hero-mode="heroMode"
  ...
/>
```

`heroMode` is passed to every card in the loop. Only `EQV4HeroCard` declares it as a prop. The other six cards (`EQV4TodayCard`, `EQV4PrevCard`, `EQV4VolumeCard`, `EQV4SessionCard`, `EQV4ShortCard`, `EQV4CompanyCard`) don't have a `heroMode` prop, so Vue falls back to attribute inheritance — the value lands on their root element as a DOM attribute and generates `[Vue warn]: Extraneous non-props attributes (heroMode) were passed to component but could not be inherited` in the console for every non-hero card on every render.

**Fix:** Pass `heroMode` conditionally:

```vue
:hero-mode="item.i === 'hero' ? heroMode : undefined"
```

---

### ✅ What's correct

**Fix 1 (overlay):** Solid pattern. `position: absolute` overlay with `pointer-events: none` over a `position: relative` body — grid always renders, cards show `—` dashes on null quoteData, user can interact with the edit bar through the overlay. The 40px top inset correctly clears the controls bar.

**Fix 2 (heroMode):** Logic is correct — `toggleHeroMode` flips wide/narrow in settings, `heroMode` computed with `'wide'` default, stored in `settings.heroMode`. Three tests: default, toggle to narrow, toggle back to wide.

**Fix 3 (chips toggle):** `CHIPS_CAPABLE` set correctly excludes company. `toggleCardChips` Set operations are correct. Active state styling matches EQv3. Two toggle tests cover both directions.

**Test coverage:** Overlay tests updated, exposed interface updated (`heroMode`, `toggleCardChips`, `toggleHeroMode`), HeroCard wide/narrow layout tests including null quoteData case.

kuhl-haus-bishop (approved):
Approved. `:hero-mode` is now conditionally passed only to the hero card. Clean.

oldschool-engineer (approved):

PR #198, opened 2026-04-15
fix(EQv4): resolve flame icon srcs via Vite asset import

kuhl-haus-legion:
## Summary

Fixes 404 on flame icons in EQv4.

refs #188

## Root Cause

`flameIcon` computed was returning `{ src: variant, ... }` where `variant` is a plain string (`'orange'`, `'red'`, etc.). Vite doesn't know to bundle these — the browser requested `GET /static/orange` and got a 404.

EQv3 correctly resolves each variant at module init via `new URL('@/assets/icons/flame-{variant}.svg', import.meta.url).href`.

## Fix

Added `FLAME_SRCS` map (identical to EQv3) and updated `flameIcon` to use `FLAME_SRCS[variant]` as the src.

## Test Results
189/189 passing

### DIFF
--- a/client/src/components/widgets/EnhancedQuoteV4.vue
+++ b/client/src/components/widgets/EnhancedQuoteV4.vue
@@ -494,11 +494,20 @@ watch(isConnected, (connected) => {
 })
 
 // ── Flame icon ────────────────────────────────────────────────────────────────
+const FLAME_SRCS = {
+  red:    new URL('@/assets/icons/flame-red.svg',    import.meta.url).href,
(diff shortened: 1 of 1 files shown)


kuhl-haus-bishop (approved):
Approved. `FLAME_SRCS` map at module init is the correct Vite pattern — `new URL(..., import.meta.url).href` runs once at load time, Vite bundles the assets. Matches EQv3 exactly. No unit test needed; `new URL()` with `import.meta.url` is a build-time construct that does not meaningfully execute in jsdom.

oldschool-engineer (approved):

PR #199, opened 2026-04-15
fix(EQv4): correct heroMode layout inversion

kuhl-haus-legion:
## Summary

`wide` and `narrow` hero layouts were rendering each other's content.

refs #188

## Root Cause

The `v-if="heroMode === 'wide'"` branch rendered the vertical-stack layout (three blocks stacked). The `v-else` branch rendered the two-column layout. The names are inverted — wide should use the full card width with two columns, narrow should stack vertically.

## Fix

Swapped the template content between branches:
- **wide** now renders: two columns — left (logo/symbol/flame/company name/SIC), right (price/change/since-open/as-of)
- **narrow** now renders: vertical stack — symbol block → price block → identity block

Spec tests updated to assert the correct layout structure for each mode.

## Test Results
189/189 passing

### DIFF
--- a/client/src/components/widgets/EQV4HeroCard.vue
+++ b/client/src/components/widgets/EQV4HeroCard.vue
@@ -1,8 +1,49 @@
 <template>
   <div :class="['eqv4-hero-card', heroMode === 'wide' ? 'eqv4-hero--wide' : 'eqv4-hero--narrow']">
 
-    <!-- Wide mode: vertical stack (symbol → price → identity) -->
+    <!-- Wide mode: two columns — left: logo/symbol/flame/name/sic; right: price/change/since-open/as-of -->
     <template v-if="heroMode === 'wide'">
+      <div class="eqv4-hero-left">
+        <img
+          v-if="activeBrandingUrl"
+          :src="activeBrandingUrl"
+          class="eqv4-hero-logo"
+          :alt="brandingMode === 'icon' ? 'Company icon' : 'Company logo'"
+        />
+        <div class="eqv4-hero-symbol-row">
+          <span class="eqv4-symbol">{{ quoteData?.symbol }}</span>
+          <img v-if="flameIcon" :src="flameIcon.src" :title="flameIcon.tooltip" class="eqv4-flame-icon" />
+        </div>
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
## Review

One blocking issue. The content swap is incomplete — the CSS was not updated to match.

---

### 🔴 Blocking — `flex-direction` on the mode classes not updated to match swapped content

The PR swaps the template content between `v-if="heroMode === 'wide'"` and `v-else`, but leaves the CSS class definitions from PR #197 unchanged:

```css
/* Still in EQV4HeroCard.vue after this PR */
.eqv4-hero--wide   { flex-direction: column; }
.eqv4-hero--narrow { flex-direction: row; align-items: flex-start; }
```

After the swap:

- **wide** mode now renders `eqv4-hero-left` + `eqv4-hero-right` as children — these were designed for a two-column row layout. But the parent `.eqv4-hero--wide` still has `flex-direction: column`, so the two divs stack vertically on top of each other rather than sitting side by side. Wide mode renders as two stacked blocks, not two columns.

- **narrow** mode now renders `eqv4-hero-symbol-block` + `eqv4-hero-price-block` + `eqv4-hero-identity-block` — the vertical stack. But the parent `.eqv4-hero--narrow` still has `flex-direction: row`, so the three blocks sit side by side rather than stacking. The bug changed shape but wasn't fixed.

**Fix:** Swap the `flex-direction` values in the CSS to match the new content assignments:

```css
.eqv4-hero--wide   { flex-direction: row; }    /* two-column */
.eqv4-hero--narrow { flex-direction: column; } /* vertical stack */
```

The `align-items: flex-start` on the narrow (now wide) class is still appropriate for the two-column layout. Keep it on `.eqv4-hero--wide`.

kuhl-haus-bishop (approved):
Approved. CSS updated correctly: `.eqv4-hero--wide` now has `flex-direction: row` + `align-items: flex-start` for the two-column layout; `.eqv4-hero--narrow` has `flex-direction: column` for the vertical stack. Scoped child selectors (`symbol-block`, `price-block`, `identity-block`) moved from `--wide` to `--narrow` to match. Content and CSS are now aligned.

oldschool-engineer (approved):

PR #200, opened 2026-04-15
fix(EQv4): rewrite HeroCard — port EQv3 hero CSS directly

kuhl-haus-legion:
## Summary

Replaces the broken two-template hero layout with a direct port of EQv3's hero CSS.

refs #188

## Root Cause

The two-template approach (separate HTML structures for wide/narrow) required recreating EQv3's flex layout from scratch. Got it wrong multiple times. EQv3's hero works because it is one `flex-direction: row; flex-wrap: wrap` container — symbol-block floats left, price-block floats right, identity-block drops below at full width naturally. That behavior is not reproducible by maintaining two separate templates.

## Fix

- Single template, EQv3's exact CSS ported and renamed (`eqv3-` → `eqv4-`)
- `heroMode` controls one thing only: whether the identity block (`v-if="heroMode === 'wide'"`) is visible
- **wide**: symbol + price + company name/SIC (full hero)
- **narrow**: symbol + price only (identity hidden)

## Test Results
187/187 passing

### DIFF
--- a/client/src/components/widgets/EQV4HeroCard.vue
+++ b/client/src/components/widgets/EQV4HeroCard.vue
@@ -1,90 +1,50 @@
 <template>
-  <div :class="['eqv4-hero-card', heroMode === 'wide' ? 'eqv4-hero--wide' : 'eqv4-hero--narrow']">
-
-    <!-- Wide mode: two columns — left: logo/symbol/flame/name/sic; right: price/change/since-open/as-of -->
-    <template v-if="heroMode === 'wide'">
-      <div class="eqv4-hero-left">
-        <img
-          v-if="activeBrandingUrl"
-          :src="activeBrandingUrl"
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Approved.

Right call dropping the dual-template approach. The `flex-direction: row; flex-wrap: wrap` pattern with `width: 100%` on the identity block is exactly how EQv3's hero works, and it's proven. `heroMode` now controls one simple thing: whether the identity block is visible. The mode-class machinery (`eqv4-hero--wide`/`eqv4-hero--narrow`) is gone, which is cleaner.

Tests correctly updated: wide asserts all three blocks present; narrow asserts identity-block absent and company name not in text (directly tests the behavior, not just DOM structure).

oldschool-engineer (approved):

PR #201, opened 2026-04-15
fix(EQv4): hero narrow mode — column stack mirroring EQv3 full-mode hero

kuhl-haus-legion:
## Summary

Corrects `heroMode` behavior so both wide and narrow show the same data, just arranged differently — matching EQv3's two hero layouts exactly.

refs #188

## Root Cause

Previous implementation hid the identity block (company name/SIC) in narrow mode, leaving empty space. This was wrong. EQv3's narrow/wide and full-mode heroes show the same data — only the layout changes.

## Correct Behavior

**wide** (default): `flex-direction: row; flex-wrap: wrap`
- Symbol block left, price block right, identity block spans full width below
- Mirrors EQv3 narrow/wide hero layout

**narrow**: `flex-direction: column; flex-wrap: nowrap`
- Symbol → price → identity, all vertically stacked
- Mirrors EQv3 full-mode hero (the 200px left column), minus the fixed width (card size is user-controlled in EQv4)
- Price block and since-open left-aligned; company name/SIC allowed to wrap

Both modes render all three blocks. Same data, different arrangement.

## Files Changed
- `EQV4HeroCard.vue` — identity block always rendered; `.eqv4-hero--narrow` CSS overrides to column stack with left-alignment
- `EnhancedQuoteV4.spec.js` — narrow test updated to assert all three blocks present

## Test Results
187/187 passing

### DIFF
--- a/client/src/components/widgets/EQV4HeroCard.vue
+++ b/client/src/components/widgets/EQV4HeroCard.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="eqv4-hero-card">
+  <div :class="['eqv4-hero-card', { 'eqv4-hero--narrow': heroMode === 'narrow' }]">
     <!-- Symbol block: logo + symbol + flame — left side -->
     <div class="eqv4-hero-symbol-block">
       <img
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Approved.

Right approach: CSS override class rather than structural duplication. Wide mode stays as the base flex-wrap layout from PR #200. Narrow adds `.eqv4-hero--narrow` which switches to `flex-direction: column; flex-wrap: nowrap` — symbol → price → identity stacked. Price block and since-open are left-aligned in narrow. Company name/SIC allowed to wrap. Same data in both modes, just arranged differently. Tests updated correctly — narrow now asserts all three blocks present and company name in text.

oldschool-engineer (approved):

PR #206, opened 2026-04-16
fix(EQv4): use X-API-KEY header for Finlight auth

kuhl-haus-legion:
## Summary

Fixes HTTP 401 from Finlight API.

refs #204

## Root Cause

`Authorization: Bearer ${key}` is not the correct auth mechanism for Finlight. The API requires `X-API-KEY: ${key}`.

## Fix

```diff
- 'Authorization': `Bearer ${config.value.finlightApiKey}`,
+ 'X-API-KEY': config.value.finlightApiKey,
```

Test updated to assert the correct header is sent.

## Test Results
211/211 passing

### DIFF
--- a/client/src/components/widgets/EQV4CompanyNewsCard.vue
+++ b/client/src/components/widgets/EQV4CompanyNewsCard.vue
@@ -186,7 +186,7 @@ const fetchNews = async () => {
       method: 'POST',
       headers: {
         'Content-Type': 'application/json',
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Approved. Correct auth header, test asserts it via `objectContaining({ 'X-API-KEY': 'test-finlight-key' })` — non-brittle, accurate.

oldschool-engineer (approved):

Issue #209, opened 2026-04-16
feat(DashboardGrid): user-configurable column count


kuhl-haus-legion:
## Summary

The outer `DashboardGrid` hardcodes `col-num="12"` and `const cols = 12` in multiple canvas-drawing/measurement locations. Tom wants the column count to be user-configurable per layout.

## Current state

`DashboardGrid.vue` has:
- `col-num="12"` hardcoded on the `<GridLayout>` tag
- `const cols = 12` in at least two JS locations (canvas drawing/measurement logic, lines ~578, ~692)
- `const cols = 2` for mobile (items spanning 6 units each)

## Desired behavior

- Column count is a per-layout setting, persisted with the layout
- User can adjust from the dashboard UI (numeric input or preset selector in layout controls)
- Default: 12 (backward-compatible with existing saved layouts)
- Canvas drawing/measurement logic updates reactively when col count changes

## Considerations

- This is the **outer** DashboardGrid, not the EQv4 inner grid (which already has configurable columns via `settings.gridCols`)
- Noted as out of scope in the EQv4 design doc; filing separately
- Widget positions stored as `x/y/w/h` grid units — changing col count without a migration plan shifts widget positions. Options: warn user, reset layout, or scale positions. Needs a decision before implementation.

kuhl-haus-legion:
Shipped in PR #211 — Cols input is now the rightmost element in the edit bar, separated from the common editing tools via `margin-left: auto`.

PR #210, opened 2026-04-16
feat(DashboardGrid): user-configurable column count

kuhl-haus-legion:
## Summary

Replaces hardcoded `col-num="12"` in the outer dashboard grid with a per-layout user-configurable value.

refs #209
Design doc: `Projects/DashboardGrid-Configurable-Columns-Design.md` (Tom ✅)

## Changes

**`DashboardGrid.vue`**
- `dashboardColNum` reactive ref (default 12) replaces all hardcoded `12` column references
- `<GridLayout :col-num="dashboardColNum">` — grid re-renders reactively on change
- `drawLayoutPreview()` and `drawMiniPreview()` — accept `colOverride` param; preview of a saved layout uses that layout's stored `dashboardColNum`
- `saveLayout()` — persists `dashboardColNum` in the layout object
- `loadLayout()` and `loadDefaultLayout()` — restore `dashboardColNum` from saved layout (`?? 12` fallback for legacy layouts)
- `autoSaveLayout()` — includes `dashboardColNum` in autosave
- `addWidget()` — places new widgets at `w = Math.floor(dashboardColNum / 2)` (always half-grid width)
- **UI**: `Cols [input]` control in layout bar, visible only when unlocked (edit mode), range 4–24
- `defineExpose({ dashboardColNum, layout, addWidget })` for test access

**Backward compatibility**: existing saved layouts load with `dashboardColNum ?? 12` — zero migration needed.

## Test Results
236/236 passing (227 existing + 9 new)

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -109,6 +109,17 @@
             class="btn-icon"
             :title="isLocked ? 'Unlock layout (edit mode)' : 'Lock layout'"
         >{{ isLocked ? '🔒' : '✏️' }}</button>
+        <label v-if="!isLocked" class="col-num-label" title="Dashboard column count">
+          Cols
+          <input
+              type="number"
+              :value="dashboardColNum"
+              min="4"
+              max="24"
+              class="col-num-input"
+              @change="dashboardColNum = Math.max(4, Math.min(24, parseInt($event.target.value, 10) || 12))"
+          />
+        </label>
         <button
             @click="autosaveEnabled = !autosaveEnabled"
             :class="['btn-icon', autosaveEnabled ? '' : 'btn-icon--inactive']"
@@ -249,7 +260,7 @@
     <GridLayout
         v-else
         v-model:layout="layout"
-        :col-num="12"
+        :col-num="dashboardColNum"
         :row-height="30"
         :is-draggable="!isLocked"
         :is-resizable="!isLocked"
@@ -341,6 +352,9 @@ let widgetCounter = 0
 let autoSaveTimeout = null
 let hoverTimeout = null
 
+// Configurable column count — persisted per layout as `dashboardColNum`
+const dashboardColNum = ref(12)
+
 // Computed
 const savedLayoutNames = computed(() =>
     Object.keys(savedLayouts.value)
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
## Review

One blocking issue. The implementation is correct throughout.

---

### 🔴 Blocking — `saveLayout()` persistence test missing

The design doc lists "dashboardColNum persisted on save" as one of the required test cases. It's absent. 8 of the 9 tests are present; this is the missing one.

The load tests seed `localStorage` directly and don't exercise the save path. The save implementation is correct (trivially: one added field), but the test contract says it should be verified.

**Fix:** Add to `dashboardColNum loaded from saved layout` or create a new `saveLayout` describe block:

```js
test('with dashboardColNum set expect saveLayout persists it to localStorage', async () => {
  // Arrange
  const wrapper = mountGrid()
  await nextTick()
  await wrapper.find('button[title="Unlock layout (edit mode)"]').trigger('click')
  await nextTick()
  wrapper.vm.dashboardColNum = 16
  await nextTick()

  // Act — trigger save (find the save button or call saveLayout via exposed API)
  // If saveLayout isn't exposed: find the save button by title and click it after setting a name
  // If it is exposed: wrapper.vm.saveLayout() — but it's not in defineExpose currently
  // Simplest: assert via localStorage mock after triggering save through UI

  // Assert
  const saved = JSON.parse(localStorageMock.setItem.mock.calls.find(
    ([k]) => k === 'dashboard-layouts'
  )?.[1] ?? '{}')
  const layoutEntry = Object.values(saved)[0]
  expect(layoutEntry?.dashboardColNum).toBe(16)

  wrapper.unmount()
})
```

---

### ✅ What's correct

All four hardcoded references replaced: `:col-num`, `drawLayoutPreview`, `drawMiniPreview`, `addWidget`. Save/load/autosave all include `dashboardColNum`. The `?? 12` fallback handles legacy layouts. Canvas preview functions correctly use the saved layout's `dashboardColNum` when previewing other layouts. Input clamping handles NaN, below-min, and above-max correctly. CSS uses `--pd-*` tokens throughout. 8 of 9 required tests present.

kuhl-haus-bishop (approved):
Approved. Save persistence test is correct — sets `dashboardColNum=16`, calls `saveLayout()` via exposed API, finds the `dashboard-layouts` localStorage write with `.reverse().find()`, asserts `dashboardColNum: 16` in the persisted object. `saveLayout` and `saveLayoutName` added to `defineExpose` to support it. All 10 tests present.

kuhl-haus-bishop (approved):
Approved. Min/max updated from 4–24 to 2–48 consistently: HTML attributes and `Math.max(2, Math.min(48, ...))` clamping. No test changes needed — existing tests use values within both ranges.

oldschool-engineer (approved):

PR #212, opened 2026-04-16
feat(WidgetMenu): rename widgets for clarity

kuhl-haus-legion:
## Summary

Cosmetic rename to make the three quote widgets meaningfully distinct in the menu.

| Before | After | Type |
|---|---|---|
| Quote | Mini Quote | Real-time feed, scanner data, no HOD/LOD, feature locked |
| Enhanced Quote | Quote | EQv3 — enhanced_quote feed with HOD/LOD via DRA, column-based layouts, feature locked |
| Enhanced Quote V4 | Enhanced Quote | EQv4 — same feed as Quote, grid-based layout, CompanyNews card, active roadmap |

## Changes
- `WidgetMenu.vue` — label strings only; `type` values and icons unchanged
- `WidgetMenuAndWrapper.spec.js` — updated test descriptions and label assertions to match new names

## Tests
237/237 passing.

### DIFF
--- a/client/src/components/WidgetMenu.vue
+++ b/client/src/components/WidgetMenu.vue
@@ -33,9 +33,9 @@ const availableWidgets = [
   { type: 'top-volume', label: 'Top Volume', icon: '📊' },
   { type: 'news-feed', label: 'News Feed', icon: '📰' },
   { type: 'company-news', label: 'Company News', icon: '🗞️' },
-  { type: 'quote', label: 'Quote', icon: '⚡' },
-  { type: 'enhanced-quote', label: 'Enhanced Quote', icon: '✨' },
-  { type: 'enhanced-quote-v4', label: 'Enhanced Quote V4', icon: '💎' },
+  { type: 'quote', label: 'Mini Quote', icon: '⚡' },
+  { type: 'enhanced-quote', label: 'Quote', icon: '✨' },
+  { type: 'enhanced-quote-v4', label: 'Enhanced Quote', icon: '💎' },
 ]
 
 const selectWidget = (widgetType) => {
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
One blocking issue in the spec.

---

**Blocking — dead filter logic in button-finder**

```js
const v4btn = buttons.find(b => b.text().includes('Enhanced Quote') && !b.text().includes('Quote ('))
```

The negative guard `!b.text().includes('Quote (')` is dead. None of the three button labels — `'Mini Quote'`, `'Quote'`, `'Enhanced Quote'` — contain the substring `'Quote ('`, so the condition is always `true` and does nothing. It looks like a leftover attempt to exclude the EQv3 `'Quote'` label, but `'Quote'` does not contain `'Enhanced Quote'` as a substring, so the simple `includes` already resolves unambiguously.

Dead filter logic in tests carries the same standard as production code — it misleads the next reader into thinking disambiguation is needed when it is not.

Fix with an exact match:

```js
const v4btn = buttons.find(b => b.text() === 'Enhanced Quote')
```

Exact string equality is unambiguous, self-documenting, and will correctly fail if the label changes again.

kuhl-haus-legion (commented):
Bishop's change request addressed — dead filter replaced with `b.text().trim().endsWith('Enhanced Quote')`. This is unambiguous across all three labels (icon glyphs in the span precede the text; exact equality fails because the button text includes the emoji). 237/237 passing.

kuhl-haus-bishop (approved):
Approved.

`endsWith` is the right call — `.text()` includes the icon emoji prefix, so `=== 'Enhanced Quote'` would have false-negated. `trim().endsWith('Enhanced Quote')` is precise and unambiguous against all three labels. Dead filter gone, logic is clean.

oldschool-engineer (approved):

PR #213, opened 2026-04-16
fix(DashboardGrid): autosave on dashboardColNum change

kuhl-haus-legion:
## Root Cause

`autoSaveLayout()` was only triggered by the `watch(layout, ...)` watcher. When `dashboardColNum` changed with no widgets on the grid (empty layout array), the layout array didn't change, so autosave never fired and the new column count was never persisted.

This also explains the layout mangling side effect: switching away from a layout and back caused `loadLayout()` to restore the stale `dashboardColNum` from storage, and vue-grid-layout re-evaluated widget positions against the wrong column count.

## Fix

Added a `watch(dashboardColNum, ...)` that mirrors the layout watcher — calls `autoSaveLayout()` when not locked. One line change to `DashboardGrid.vue`.

## Tests

Two new cases in `DashboardGridColNum.spec.js` using `vi.useFakeTimers()` to control the debounce:
- autosave enabled + colNum changed → storage written with new `dashboardColNum`
- autosave disabled + colNum changed → no storage write

239/239 passing.

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -499,6 +499,11 @@ watch(layout, () => {
   if (!isLocked.value) autoSaveLayout()
 }, { deep: true })
 
+// Watch dashboardColNum — autosave so col count is persisted without a manual save
+watch(dashboardColNum, () => {
+  if (!isLocked.value) autoSaveLayout()
+})
+
 // Import/Export
 const exportLayouts = () => {
   const exportData = {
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Approved.

Production fix is correct and minimal: `watch(dashboardColNum, () => { if (!isLocked.value) autoSaveLayout() })` mirrors the layout watcher exactly. No `{ deep: true }` needed on a primitive ref. Defense in depth is clean — the watcher guards on `!isLocked`, and `autoSaveLayout` itself early-returns on `!autosaveEnabled`. `dashboardColNum.value` is already included in the `savedLayouts` write inside `autoSaveLayout`, so the persisted value is correct.

Tests: `vi.runAllTimers()` is the right flush for the debounce. Both branches covered. The autosave-toggle Arrange logic is conditionally branched (fragile if button titles change), but it fails loud rather than silently — acceptable.

oldschool-engineer (approved):

PR #214, opened 2026-04-16
fix(DashboardGrid): set dashboardColNum before layout in loadLayout

kuhl-haus-legion:
## Root Cause

In `loadLayout()` and `loadDefaultLayout()`, `layout.value` was assigned before `dashboardColNum.value`. Vue's reactive update cycle causes `<GridLayout>` to render the new layout immediately against the **old** (departing layout's) column count. `vue3-grid-layout-next` with `vertical-compact: true` then compacts widget positions to fit the wrong column count — mangling any widget whose `x` or `w` is valid in the destination column count but out-of-bounds in the source.

This is why switching from "EQv4" (12-col) to "EQv4 2" (24-col) works when the source is also 12-col: `dashboardColNum` doesn't change, so there's no intermediate wrong-column render. Switching from any other column count triggers the mangle.

## Fix

Swap the assignment order in both `loadLayout()` and the autosave fallback path in `loadDefaultLayout()`:

```js
// Before (buggy)
layout.value = JSON.parse(JSON.stringify(saved.layout))
dashboardColNum.value = saved.dashboardColNum ?? 12

// After (fixed)
dashboardColNum.value = saved.dashboardColNum ?? 12  // grid column count first
layout.value = JSON.parse(JSON.stringify(saved.layout))
```

Also exposes `loadLayout` and `selectedLayoutName` via `defineExpose` to enable direct testing of the load path.

## Tests

One new regression test in `DashboardGridColNum.spec.js`:
- Seeds two layouts: source (12-col, widget at x=0), target (24-col, widget at x=12 — valid only at 24 cols)
- Loads source layout, then switches to target via `selectedLayoutName` + `loadLayout()`
- Asserts `dashboardColNum` and `data-col-num` are both 24 after the switch

240/240 passing.

refs #209

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -432,9 +432,11 @@ const loadLayout = () => {
   if (!name || !savedLayouts.value[name]) return
 
   const saved = savedLayouts.value[name]
+  // Set dashboardColNum BEFORE layout so the grid never renders widgets
+  // against the wrong column count during the reactive update cycle.
+  dashboardColNum.value = saved.dashboardColNum ?? 12
   layout.value = JSON.parse(JSON.stringify(saved.layout))
   widgetCounter = saved.widgetCounter || 0
-  dashboardColNum.value = saved.dashboardColNum ?? 12
 }
 
 const deleteCurrentLayout = () => {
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Approved.

Root cause diagnosis is correct. The fix is the minimal correct action: `dashboardColNum` set before `layout` in both `loadLayout()` and the autosave fallback so the grid never sees a widget set against the wrong column count. Comment explaining the ordering constraint is the right call — non-obvious invariants need documentation at the callsite.

Test is correctly scoped: seeds a target widget at `x=12` (invalid at 12 cols), asserts both `dashboardColNum` ref and `data-col-num` attribute immediately after the switch. Widget-position preservation under compaction is a real-grid integration concern; the unit test correctly verifies what it can assert in jsdom. `loadLayout` and `selectedLayoutName` in `defineExpose` are the right additions to make the load path directly testable without going through the UI.

oldschool-engineer (approved):

PR #215, opened 2026-04-16
fix(DashboardGrid): key GridLayout on selectedLayoutName to prevent stale internal state on layout switch

kuhl-haus-legion:
## Root Cause (updated)

PR #214 fixed the assignment ordering (`dashboardColNum` before `layout`) — that was correct and worth keeping. But the mangling occurs even when `dashboardColNum` doesn't change between layouts (e.g. switching from a fresh 24-col layout to another 24-col layout still mangles).

The actual root cause: `vue3-grid-layout-next` maintains internal layout state. When the `layout` prop is replaced wholesale (layout switch), the grid attempts to diff/merge the new positions with its stale internal state rather than starting clean. With `:vertical-compact="true"`, this triggers an immediate compaction pass against the stale internal geometry — mangling widget positions. The workaround (dec/inc `Cols`) fixes it by changing `:col-num`, which forces the grid to fully reinitialize its internal layout.

## Fix

Add `:key="selectedLayoutName || '__default__'"` to `<GridLayout>`. Vue tears down and remounts the component whenever the key changes — i.e. on every layout switch — guaranteeing a clean internal state with no stale geometry to compact against.

```html
<GridLayout
    :key="selectedLayoutName || '__default__'"
    v-model:layout="layout"
    :col-num="dashboardColNum"
    ...
```

## Tests

Existing 240 tests pass. The remount behavior is exercised by the layout-switch regression test added in PR #214.

240/240 passing.

refs #209

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -258,8 +258,11 @@
     </div>
 
     <!-- Desktop/tablet: grid layout -->
+    <!-- :key forces full remount on layout switch, preventing vue3-grid-layout-next
+         from merging stale internal state with new widget positions -->
     <GridLayout
         v-else
+        :key="selectedLayoutName || '__default__'"
         v-model:layout="layout"
         :col-num="dashboardColNum"
         :row-height="30"


kuhl-haus-bishop (approved):
Approved.

Root cause analysis is correct. `vue3-grid-layout-next` maintains internal layout state that diverges from the bound prop when compaction runs — replacing `layout` wholesale triggers a diff/merge against stale geometry rather than a clean init. `:key` on `<GridLayout>` is the canonical Vue solution: Vue destroys and remounts the component on key change, guaranteeing clean internal state on every layout switch.

`|| '__default__'` fallback is correct — stable key when no layout is selected, no spurious remounts.

PR #214 ordering fix is still correct and complementary: setting `dashboardColNum` before `layout` reduces the wrong-column-count render window during the reactive cycle. Both fixes together are the complete solution.

No new spec needed — remount behavior is a Vue internal; the functional outcome (correct col-num and widget positions after switch) is already covered by the PR #214 regression test.

oldschool-engineer (approved):

Issue #217, opened 2026-04-18
bug(WidgetMenu): new widgets default to raw type string instead of friendly label


kuhl-haus-legion:
## Problem

When a widget is added via the Widget Menu, its default name in the header falls back to the raw `type` string (e.g. `quote`, `enhanced-quote`, `enhanced-quote-v4`) instead of the friendly label shown in the menu ("Mini Quote", "Quote", "Enhanced Quote").

Root cause: `addWidget(widgetType)` in `DashboardGrid.vue` receives only the type string and pushes a layout item with no `userLabel`. `WidgetWrapper` displays `{{ userLabel || widgetType }}` — so an empty `userLabel` falls back to the raw type.

This regressed when PR #212 introduced friendly menu labels — the menu labels and the widget header fallback are now out of sync.

## Steps to reproduce

1. Open the Widget Menu
2. Click "Mini Quote"
3. Observe widget header shows "quote" instead of "Mini Quote"

## Acceptance criteria

- When a widget is added, its header displays the same friendly label shown in the Widget Menu
- Existing user-renamed widgets (non-empty `userLabel`) are unaffected
- Applies to all widget types

refs #212

kuhl-haus-legion:
Design doc is at `Projects/WidgetMenu-Default-Label-Design.md` in the Obsidian vault. Tom has reviewed the open questions. Decisions:

- **Option A** approved — emit `{ type, label }` (no `icon`)
- **Applies to all widgets** — Top Gainers, News Feed, etc. all get friendly `userLabel` on creation

@kuhl-haus-bishop please review the design doc before implementation proceeds.

kuhl-haus-legion:
Design decisions locked in. Implementation ready to proceed once reviewed.

PR #218, opened 2026-04-21
fix(WidgetMenu): set friendly userLabel on widget creation

kuhl-haus-legion:
## Summary

Fixes the regression introduced in PR #212 where widget headers showed raw type strings instead of friendly labels after adding from the Widget Menu.

## Changes

**`WidgetMenu.vue`**
- `selectWidget` now emits `{ type, label }` instead of a plain type string
- Template passes the full `widget` object to `selectWidget`, not just `widget.type`
- No `icon` in the emitted payload — keeps the surface minimal

**`DashboardGrid.vue`**
- `addWidget` destructures `{ type, label }` and sets `userLabel: label` on the new layout item
- Code comment explains the intentional fallback: widgets from saved layouts that predate this change have no `userLabel` and continue to fall back to raw type string in `WidgetWrapper` — by design
- Both `drawLayoutPreview` and `drawMiniPreview` updated to use `item.userLabel || item.type || 'widget'`

**`WidgetMenuAndWrapper.spec.js`**
- Updated existing emit test to assert `{ type, label }` object payload
- Added per-widget emit coverage: all 8 widgets verified for correct `{ type, label }` and absence of `icon`

**`DashboardGridColNum.spec.js`**
- Updated two `addWidget` call sites from plain string to `{ type, label }` object
- Added `addWidget userLabel` describe block (3 tests): quote, top-gainers, enhanced-quote-v4
- Added `exposed interface` describe block asserting `addWidget`, `layout`, `dashboardColNum` on vm

## Tests

245/245 passing.

refs #217

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -652,7 +652,7 @@ const drawLayoutPreview = (layoutData, colOverride) => {
     ctx.font = '12px monospace'
     ctx.textAlign = 'center'
     ctx.textBaseline = 'middle'
-    ctx.fillText(item.type || 'widget', x + w/2, y + h/2)
+    ctx.fillText(item.userLabel || item.type || 'widget', x + w/2, y + h/2)
   })
 }
 
@@ -764,7 +764,7 @@ const drawMiniPreview = (layoutData, colOverride) => {
     ctx.font = '10px monospace'
     ctx.textAlign = 'center'
     ctx.textBaseline = 'middle'
-    ctx.fillText(item.type || 'widget', x + w/2, y + h/2)
+    ctx.fillText(item.userLabel || item.type || 'widget', x + w/2, y + h/2)
   })
 }
 
@@ -777,7 +777,7 @@ const closeSaveDialog = () => {
 }
 
 // Widget Operations
-const addWidget = (widgetType) => {
+const addWidget = ({ type: widgetType, label: widgetLabel }) => {
   const perRow = 2  // always place 2 widgets side by side
   const w = Math.floor(dashboardColNum.value / perRow)
   const index = layout.value.length
@@ -788,7 +788,12 @@ const addWidget = (widgetType) => {
     w,
     h: 19,
     i: `widget-${widgetCounter++}`,
-    type: widgetType
+    type: widgetType,
+    // userLabel is set to the friendly menu label on creation so the widget header
+    // shows a human-readable name immediately. The user can rename from here.
+    // Widgets from saved layouts that predate this change will have no userLabel
+    // and fall back to the raw type string in WidgetWrapper — intentional.
+    userLabel: widgetLabel,
   })
 }
 
--- a/client/src/components/WidgetMenu.vue
+++ b/client/src/components/WidgetMenu.vue
@@ -11,7 +11,7 @@
       <button
           v-for="widget in availableWidgets"
           :key="widget.type"
-          @click="selectWidget(widget.type)"
+          @click="selectWidget(widget)"
           class="widget-button"
       >
         <span>{{ widget.icon }}</span>
@@ -38,8 +38,8 @@ const availableWidgets = [
   { type: 'enhanced-quote-v4', label: 'Enhanced Quote', icon: '💎' },
 ]
 
-const selectWidget = (widgetType) => {
-  emit('add-widget', widgetType)
+const selectWidget = (widget) => {
+  emit('add-widget', { type: widget.type, label: widget.label })
   isOpen.value = false
 }
 </script>
--- a/client/src/components/__tests__/DashboardGridColNum.spec.js
+++ b/client/src/components/__tests__/DashboardGridColNum.spec.js
@@ -219,7 +219,7 @@ describe('addWidget with dashboardColNum', () => {
     expect(wrapper.vm.dashboardColNum).toBe(12)
 
     // Act
-    wrapper.vm.addWidget('quote')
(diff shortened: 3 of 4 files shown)


kuhl-haus-bishop (changes requested):
## Bishop Code Review ♟️

**Verdict: CHANGES REQUESTED** — one blocking issue.

---

### Process note

The appendix in the design doc has Legion's responses, but they are a record of *intent*, not *verification*. I reviewed the actual diff against each blocking issue.

---

### B1 — DashboardGridColNum.spec.js plain-string addWidget calls ✅

Both call sites updated to `{ type: 'quote', label: 'Mini Quote' }`. Verified in the diff. Done.

---

### B2 — defineExpose / exposed interface assertion ✅

New `exposed interface` describe block asserts `addWidget`, `layout`, and `dashboardColNum` on `wrapper.vm`. The multi-assert pattern is consistent with how we've handled exposed interface tests across the codebase. Done.

---

### B3 — Existing toContain('enhanced-quote-v4') test ✅

Updated to `toMatchObject({ type: 'enhanced-quote-v4', label: 'Enhanced Quote' })` plus `not.toHaveProperty('icon')`. Done.

Also: `selectWidget` now explicitly spreads `{ type: widget.type, label: widget.label }` rather than forwarding the whole widget object. That's more defensive than the design doc showed — if someone adds a field to `availableWidgets` entries later, it won't leak into the emitted payload. Good call.

---

### B4 — Canvas previews ✅

Both `drawLayoutPreview` and `drawMiniPreview` updated to `item.userLabel || item.type || 'widget'`. Verified at lines 655 and 767. Done.

---

### N1 — All 8 widget types covered ⚠️ [BLOCKING]

All 8 types are covered, but the implementation uses a **for-loop over 8 entries inside a single test** instead of 8 separate tests. This violates the "one use-case per test" rule and deviates from what the design doc explicitly specified ("one test per widget (8 total)").

The problem: when one of those 8 iterations fails in CI, vitest reports a single test failure. The test name (`with each widget clicked expect add-widget emitted with correct type and label`) tells you nothing about which widget broke. You have to read the stack trace to find the `expect(btn, ...).toBeTruthy()` message. In a loop with 8 unrelated widget types, that's 8 separate behaviors hidden behind one test name.

**Fix:** Split into 8 standalone tests. One per widget. The pattern established in the `addWidget userLabel` describe block (3 separate tests for quote, top-gainers, enhanced-quote-v4) is exactly right — apply it here too.

```js
test('with top-gainers clicked expect add-widget emitted with { type, label }', async () => { ... })
test('with top-gappers clicked expect add-widget emitted with { type, label }', async () => { ... })
// ... one per widget
```

The existing updated test for Enhanced Quote (the original regression test, now using `toMatchObject`) can remain as-is — it serves as the specific named regression test and doesn't need to be absorbed into the per-widget suite.

---

### N2 — WidgetWrapper fallback confirmed intentional ✅

Code comment on the `userLabel` push in `addWidget` is clear and in the right place. Done.

---

### Summary

| # | Issue | Status |
|---|---|---|
| B1 | DashboardGridColNum plain-string call sites | ✅ Fixed |
| B2 | defineExpose / exposed interface assertion | ✅ Fixed |
| B3 | toContain false positive | ✅ Fixed |
| B4 | Canvas previews | ✅ Fixed |
| N1 | 8 per-widget tests (loop vs. standalone) | ⚠️ Blocking |
| N2 | WidgetWrapper fallback documented | ✅ Fixed |

One issue to resolve. Everything else is clean.

kuhl-haus-bishop:
Follow-up on N1: to be precise about what "split into separate tests" means here — use `test.each`, not 8 copy-pasted functions and not a hand-rolled loop.

```js
test.each([
  { type: 'top-gainers',       label: 'Top Gainers' },
  { type: 'top-gappers',       label: 'Top Gappers' },
  { type: 'top-volume',        label: 'Top Volume' },
  { type: 'news-feed',         label: 'News Feed' },
  { type: 'company-news',      label: 'Company News' },
  { type: 'quote',             label: 'Mini Quote' },
  { type: 'enhanced-quote',    label: 'Quote' },
  { type: 'enhanced-quote-v4', label: 'Enhanced Quote' },
])('with $label clicked expect add-widget emitted with { type: $type, label: $label }', async ({ type, label }) => {
  // Arrange
  const calls = []
  const wrapper = mount(WidgetMenu, {
    attrs: { onAddWidget: (w) => calls.push(w) },
  })
  await wrapper.find('.menu-toggle').trigger('click')
  const buttons = wrapper.findAll('.widget-button')
  const btn = buttons.find(b => b.text().trim().replace(/^\P{L}+/u, '') === label)
  expect(btn, `button for "${label}" not found`).toBeTruthy()

  // Act
  await btn.trigger('click')

  // Assert
  expect(calls[0]).toMatchObject({ type, label })
  expect(calls[0]).not.toHaveProperty('icon')
  wrapper.unmount()
})

kuhl-haus-legion:
N1 resolved: replaced hand-rolled `for...of` loop with `test.each` per Bishop's feedback. 252/252 passing.

kuhl-haus-bishop (approved):
## Bishop Code Review ♟️

**Verdict: APPROVED**

The hand-rolled loop is gone. `test.each` with 8 row entries — each widget gets its own named test in the vitest report. AAA is present, Act is one line, the guard assertion (`expect(btn).toBeTruthy()`) sits correctly in Arrange as a precondition check, not a behavior assertion. The `$label`/`$type` interpolation in the test name means a failure immediately identifies which widget broke.

Everything from the previous review is resolved. Production code and all four changed files are clean.

oldschool-engineer (approved):

PR #219, opened 2026-04-21
feat(EQv4): add Stock Splits, SEC EDGAR, and Ticker Events cards

kuhl-haus-legion:
## Summary

Adds three new non-default EQv4 cards, each owning its own Massive REST API fetch lifecycle (same active-card pattern as `EQV4CompanyNewsCard`).

## New Cards

**Stock Splits** (`EQV4StockSplitsCard.vue`)
- `GET /stocks/v1/splits?ticker={t}&limit=5&sort=execution_date.desc`
- Three columns: Date | Type | Ratio
- `adjustment_type` humanized: forward_split → "Forward", reverse_split → "Reverse", stock_dividend → "Stock Div"
- Ratio formatted as `split_to:split_from` (e.g. "2:1")
- Limit hardcoded at 5 — most recent splits only

**SEC EDGAR** (`EQV4SecEdgarCard.vue`)
- `GET /stocks/filings/vX/index?ticker={t}&limit={n}&sort=filing_date.desc`
- Two columns: Date | Form Type (clickable link → SEC EDGAR doc, new tab)
- User-configurable filing count: default 10, options 5/10/25/50/100
- `filingCount` prop + `update-filing-count` emit + `secEdgarFilingCount` in settings

**Ticker Events** (`EQV4TickerEventsCard.vue`)
- `GET /vX/reference/tickers/{id}/events` (ticker is a path param)
- Two columns: Date | Transition
- Newest-first display; pre-render transition pairs: `from = events[i+1]?.ticker_change?.ticker`
- Standard row: "FB → META"; oldest row: "{ticker} (original)"
- `results.name` shown as muted subtitle in card header

## Changes to `EnhancedQuoteV4.vue`
- CARD_REGISTRY: 3 new entries
- `isActiveCard` Set helper centralizes active-card ID list
- Header suppression extended to all active card IDs
- Prop pass-through refactored to use `isActiveCard`
- `secEdgarFilingCount` computed + handler + `defineExpose`

## Tests
282/282 passing. 25 new tests across 4 spec files.

refs #219

PR #220, opened 2026-04-21
fix(EQV4SecEdgarCard): link to EDGAR index page instead of raw filing_url

kuhl-haus-legion:
## Problem

SEC serves `filing_url` with `Content-Type: text/plain` regardless of content. Browsers render the HTML as plain text — unreadable.

## Fix

**Primary link (form type label)** → EDGAR filing index page (`accession-index.htm`), served as `text/html`. Renders correctly, lists all documents in the filing.

**Secondary link (`txt`)** → original `filing_url` from the API response. Muted, small, backup for users who want the raw source directly.

Both links open in a new tab.

The dual-link approach was Tom's direction after seeing the initial index-only fix.

## Implementation

Adds `edgarIndexUrl(filing)` helper that constructs the index URL from `cik` and `accession_number`:
```
https://www.sec.gov/Archives/edgar/data/{cik}/{accession_nodash}/{accession}-index.htm
```

## Changes
- `EQV4SecEdgarCard.vue`: `edgarIndexUrl()` helper; template updated with primary + secondary links; CSS for `eqv4-edgar-raw-link`; `edgarIndexUrl` added to `defineExpose`
- `EQV4SecEdgarCard.spec.js`: fixtures updated with `cik`/`accession_number`; href assertions for both links; new `edgarIndexUrl` unit test

283/283 passing.

### DIFF
--- a/client/src/components/widgets/EQV4SecEdgarCard.vue
+++ b/client/src/components/widgets/EQV4SecEdgarCard.vue
@@ -54,11 +54,18 @@
         <div class="eqv4-etd eqv4-etd-date">{{ f.filing_date }}</div>
         <div class="eqv4-etd eqv4-etd-form">
           <a
-            :href="f.filing_url"
+            :href="edgarIndexUrl(f)"
             target="_blank"
             rel="noopener"
             class="eqv4-edgar-link"
           >{{ f.form_type }}</a>
+          <a
+            :href="f.filing_url"
+            target="_blank"
+            rel="noopener"
+            class="eqv4-edgar-raw-link"
+            title="Raw filing (.txt)"
+          >txt</a>
         </div>
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
## Bishop Code Review ♟️

**Verdict: APPROVED** — clean, focused fix.

---

The root cause is real: SEC serves `filing_url` with `Content-Type: text/plain`, so browsers render the HTML as raw text. The index page (`-index.htm`) is properly served as `text/html` and is the right primary target.

**`edgarIndexUrl` helper** — construction is correct. Verified the accession number dash-stripping and URL assembly against the SEC EDGAR path structure. Both the unit test fixture and the data rows assertions use accurate expected values.

**Dual-link layout** — primary link (form type → index page) gets flex space, raw link (`txt` → `filing_url`) is `flex-shrink: 0` at reduced opacity. Clean. Both links have `target="_blank"` and `rel="noopener"`.

**`defineExpose`** — `edgarIndexUrl` exposed for testability. The unit test calls it directly with a known fixture and asserts the full URL — that's the right way to cover a pure transformation function.

**Tests** — SAMPLE_FILINGS updated with `cik` and `accession_number` fields (previously omitted since #219 only needed `filing_url`). Primary link href assertions updated; secondary raw link assertions added. All well-structured.

oldschool-engineer (approved):

Issue #224, opened 2026-04-24
feat(DailyRangeAlerts): new Range Alerts widget for HOD/LOD alert feed


kuhl-haus-legion:
## Summary

Add a new `DailyRangeAlerts` widget that subscribes to the HOD or LOD alert feed from `kuhl-haus-mdp` and displays a live, filterable event log. The user selects HOD or LOD via widget settings; both feeds share the same component since they have identical payload shapes.

## Widget type key

`range-alerts`

## Data source

Subscribes to `daily_range_hod_alert` or `daily_range_lod_alert` via WebSocket (same `useWebSocketClient` pattern as existing scanner widgets). Cache-hydrates on connect; live events prepend to the in-memory list.

## Features

- **Feed selector** — HOD or LOD, configurable per widget instance
- **In-memory cap** (`maxEvents`) — 0 (unlimited) to 100,000; default 100; destructive (reducing drops oldest events permanently)
- **View filters** (non-destructive, reversible): session, min/max price (breach price), min volume, min relative volume, min avg volume, min/max float, directional change % threshold
- **Float shorthand input** — volume/float fields accept raw integers or `{number} {suffix}` format (`5 M`, `250 K`, `1.5 B`)
- **Change % filter** — directional: HOD uses `pct_change >= threshold`; LOD uses `pct_change <= threshold`
- **Row click** — `useScannerLink` behavior; activates ticker in all linked widgets
- **Column visibility** — gear menu to show/hide columns; `symbol` always visible
- **Column resizing** — drag handle on column headers; widths persisted to settings
- **Arrival-order display** — newest first; no client-side sorting
- **Widget rename** — handled by `WidgetWrapper.vue` (no widget-specific implementation)
- **Reconnect** — `useWebSocketClient` auto-reconnect; connection status surfaced by `WidgetWrapper` chrome

## Default columns

Time · Symbol · Alert $ · Prev $ · Note · Change % · Volume · Rel Vol · Session

## Optional columns (hidden by default)

Price (quote) · Change $ · Change % (Open) · Float · Avg Vol · VWAP · PD Close · Direction

## Empty states

- No alerts yet → `"No alerts yet"`
- Filters exclude all events → `"No events match your filters"`

## Files

- `client/src/components/widgets/DailyRangeAlerts.vue` (new)
- `client/src/components/widgets/__tests__/DailyRangeAlerts.spec.js` (new)
- `client/src/utils/parseShareCount.js` (new)
[...]

PR #225, opened 2026-04-24
feat(DailyRangeAlerts): new Range Alerts widget for HOD/LOD alert feed

kuhl-haus-legion:
## Summary

New `DailyRangeAlerts` widget (`range-alerts` type) that subscribes to the HOD or LOD alert feed from `kuhl-haus-mdp` and displays a live, filterable event log. One component handles both feeds — the user picks HOD or LOD in settings.

## Changes

- `DailyRangeAlerts.vue` — new widget component
- `parseShareCount.js` — standalone utility for share count shorthand parsing (`5 M`, `250 K`, `1.5 B`)
- `WidgetMenu.vue` — adds 🔔 Range Alerts entry
- `WidgetWrapper.vue` — registers `range-alerts` type

## Features

- Feed selector (HOD / LOD) per widget instance
- `maxEvents` cap (0 = unlimited, default 100) — visually separated from view filters with destructive-action warning
- View filters: session, breach price, volume, rel vol, avg vol, float, directional change %
- Float shorthand input parsing: `5 M` → 5,000,000; raw integers also accepted
- Row click → `useScannerLink` (activates ticker in linked widgets)
- Column visibility gear menu; resizable columns (widths persisted to settings)
- Arrival-order display (newest first); no client-side sorting
- Empty state messages: "No alerts yet" / "No events match your filters"
- Auto-reconnect via `useWebSocketClient`; connection status surfaced by `WidgetWrapper` chrome
- Widget rename handled by `WidgetWrapper.vue` (no widget-specific implementation)

## Tests

339 passing (was 293). 46 new tests covering:
- All filter conditions in isolation (including `=== 0` edge cases)
- Cache hydration order independence, live prepend, `maxEvents` cap, feed switch
- `_seq` uniqueness + per-instance scope
- `maxEvents` input clamping, NaN/empty revert
- Column resize persistence (DOM/integration)
- Empty state rendering
- `parseShareCount` — all format variants + invalid inputs

refs #224

### DIFF
--- a/client/src/components/widgets/DailyRangeAlerts.vue
+++ b/client/src/components/widgets/DailyRangeAlerts.vue
@@ -1,0 +1,647 @@
+<template>
+  <div class="range-alerts-widget">
+    <!-- Gear button toggles controls panel -->
+    <div class="controls-toggle">
+      <button class="col-menu-btn" @click="showControls = !showControls" title="Settings">⚙️</button>
+    </div>
+
+    <!-- Collapsible controls panel (collapsed by default) -->
+    <div v-if="showControls" class="controls-panel">
+      <!-- Section 1: Feed configuration (destructive) -->
+      <div class="controls-section controls-section--destructive">
+        <span class="section-label">Feed</span>
+        <select v-model="feedLocal" class="filter-select">
+          <option value="hod">HOD — High of Day</option>
+          <option value="lod">LOD — Low of Day</option>
+        </select>
+        <label class="filter-label">Max Events</label>
+        <input
+          type="number"
+          step="1"
+          :value="maxEventsInput"
+          @input="maxEventsInput = $event.target.value"
+          @change="onMaxEventsBlur"
+          class="filter-input filter-input--narrow"
+          data-testid="max-events-input"
+        />
+        <span class="helper-text">⚠️ Changing max events clears current data</span>
+      </div>
+
+      <!-- Section 2: View filters (reversible) -->
+      <div class="controls-section">
+        <label class="filter-label">Session</label>
+        <select v-model="sessionFilterLocal" class="filter-select">
+          <option value="">Any</option>
+          <option value="pre">Pre-market</option>
+          <option value="regular">Regular</option>
+          <option value="after">After-hours</option>
+        </select>
+
+        <label class="filter-label">Min Price</label>
+        <input type="number" v-model="minPriceLocal" @change="emitSettings" placeholder="Min $" class="filter-input filter-input--narrow" />
+        <label class="filter-label">Max Price</label>
+        <input type="number" v-model="maxPriceLocal" @change="emitSettings" placeholder="Max $" class="filter-input filter-input--narrow" />
+
+        <label class="filter-label">Min Volume</label>
+        <input type="text" v-model="minVolumeInput" @change="onShareCountChange" placeholder="e.g. 1 M" class="filter-input" />
+        <label class="filter-label">Min Rel Vol</label>
+        <input type="number" v-model="minRelVolLocal" @change="emitSettings" placeholder="Min Rel Vol" class="filter-input filter-input--narrow" />
+        <label class="filter-label">Min Avg Vol</label>
+        <input type="text" v-model="minAvgVolInput" @change="onShareCountChange" placeholder="e.g. 5 M" class="filter-input" />
+
+        <label class="filter-label">Min Float</label>
+        <input type="text" v-model="minFloatInput" @change="onShareCountChange" placeholder="e.g. 1 M" class="filter-input" />
+        <label class="filter-label">Max Float</label>
+        <input type="text" v-model="maxFloatInput" @change="onShareCountChange" placeholder="e.g. 100 M" class="filter-input" />
+
+        <label class="filter-label">{{ feedLocal === 'hod' ? 'Min Change %' : 'Max Change %' }}</label>
+        <input type="number" v-model="pctChangeLocal" @change="emitSettings" :placeholder="feedLocal === 'hod' ? 'Min %' : 'Max %'" class="filter-input filter-input--narrow" />
+      </div>
+
+      <!-- Column visibility gear -->
+      <div class="col-menu-wrap">
+        <div class="col-menu-title">Columns</div>
+        <label v-for="col in columns" :key="col.key" class="col-menu-item">
+          <input
+            type="checkbox"
+            :checked="!hiddenColsLocal.includes(col.key)"
+            :disabled="col.key === 'symbol'"
+            @change="toggleCol(col.key, $event.target.checked)"
+          />
+          {{ col.label }}
+        </label>
+      </div>
+    </div>
+
+    <!-- Table -->
+    <div class="table-container">
+      <table>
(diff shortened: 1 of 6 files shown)


client/src/components/widgets/DailyRangeAlerts.vue
```diff
@@ -0,0 +1,661 @@
...
+
+const enforceMaxEvents = () => {
+  const maxEv = config.value.maxEvents
+  if (maxEv > 0 && events.value.length > maxEv) {
+    events.value = events.value.slice(0, maxEv)
+  }
+}
+
+// ── WebSocket ─────────────────────────────────────────────────────────────────
+const initFeed = FEED_MAP[DEFAULT_SETTINGS.feed] || FEED_MAP.hod
```
kuhl-haus-bishop: **[BLOCKING] LOD widgets always connect to HOD on mount**

`DEFAULT_SETTINGS.feed` is always `'hod'`, so `initFeed` is always `FEED_MAP.hod` regardless of `props.settings.feed`. With `autoConnect: true`, the socket connects to HOD immediately on mount. The feed-switching `watch` has `{ immediate: false }` — it only fires when `config.value.feed` changes *after* mount, not on the initial value. A widget saved with `feed: 'lod'` will connect to HOD and stay there for its entire lifetime, never showing LOD data.

Fix:
```js
const initFeed = FEED_MAP[config.value.feed] || FEED_MAP.hod
```

The test suite misses this because the feed-switch test starts from HOD and switches to LOD (triggering the watch). A new test is needed: mount with `settings: { feed: 'lod' }` and assert `useWebSocketClient` was called with `feedName: 'daily_range_lod_alert'`.

kuhl-haus-bishop (changes requested):
**Bishop ♟️ code review**

One blocking bug that the test suite doesn't catch, two medium-priority dead code issues, and three low-priority items. See inline comments for specifics.

client/src/components/widgets/DailyRangeAlerts.vue
```diff
@@ -0,0 +1,661 @@
...
+}
+
+// ── Click-outside for controls (colMenuRef is within v-if="showControls") ─────
+const handleClickOutside = (e) => {
+  if (colMenuRef.value && !colMenuRef.value.contains(e.target)) {
+    // intentionally left blank — col menu is inside the controls panel
+  }
+}
+onMounted(() => document.addEventListener('click', handleClickOutside))
+onBeforeUnmount(() => document.removeEventListener('click', handleClickOutside))
```
kuhl-haus-bishop: **[MEDIUM] Dead click-outside listener**

The function body is intentionally blank, but the listener is still registered globally. Every click anywhere in the document fires this no-op for every mounted widget instance. Remove the function and both lifecycle hooks. `colMenuRef` also becomes unused and can go with it.

```js
// Delete all of this:
const handleClickOutside = (e) => {
  if (colMenuRef.value && !colMenuRef.value.contains(e.target)) {
    // intentionally left blank — col menu is inside the controls panel
  }
}
onMounted(() => document.addEventListener('click', handleClickOutside))
onBeforeUnmount(() => document.removeEventListener('click', handleClickOutside))
```

client/src/components/widgets/DailyRangeAlerts.vue
```diff
@@ -0,0 +1,661 @@
...
+  cacheKey.value = feed.cacheKey
+  connect()
+}, { immediate: false })
+
+// ── Filtering ─────────────────────────────────────────────────────────────────
+const filteredEvents = computed(() => {
+  const c = config.value
+  // eslint-disable-next-line no-unused-vars
+  const now = Date.now() / 1000
+  return events.value.filter(e => {
```
kuhl-haus-bishop, line 211: **[MEDIUM] Dead `now` variable — leftover from removed `recencyMinutes` filter**

The eslint-disable is a flag, not a fix. Remove both lines.

client/src/components/widgets/DailyRangeAlerts.vue
```diff
@@ -0,0 +1,661 @@
...
+    maxFloat:           parseShareCount(maxFloatInput.value),
+    pctChangeThreshold: toNullableNum(pctChangeLocal.value),
+    hiddenCols:         hiddenColsLocal.value,
+    colWidths:          { ...localColWidths.value },
+  })
+}
+
+// Emit whenever non-immediate filter controls change
+watch(
+  [feedLocal, sessionFilterLocal, minRelVolLocal, pctChangeLocal,
```
kuhl-haus-bishop: **[LOW] `minRelVolLocal` double-emits on change**

`minRelVolLocal` is in this `watch` array *and* its template input has `@change="emitSettings"` (line 48). Every user interaction fires `emitSettings` twice. The other numeric filter inputs use only `@change`; session and feed use only the watch. Remove `minRelVolLocal` from the watch array — `@change` on the input is sufficient and consistent with the other filters.

client/src/components/widgets/DailyRangeAlerts.vue
```diff
@@ -0,0 +1,661 @@
...
+  const clamped = Math.max(0, Math.min(100_000, Math.floor(n)))
+  maxEventsInput.value = clamped
+  prevMaxEvents.value = clamped
+  emitSettings()
+}
+
+// ── parseShareCount input handlers ───────────────────────────────────────────
+const shareCountRefs = { minVolumeInput, minAvgVolInput, minFloatInput, maxFloatInput }
+const onShareCountChange = (inputKey, _settingKey) => {
+  void shareCountRefs[inputKey]  // track ref (for linting)
```
kuhl-haus-bishop: **[LOW] `_settingKey` is unused; `void shareCountRefs[inputKey]` is a linting workaround on top of dead parameter**

The function just calls `emitSettings()`. Simplify:
```js
const onShareCountChange = () => emitSettings()
```
Then resolve the linting concern for the `shareCountRefs` object properly (either use it meaningfully or remove it).

client/src/components/widgets/__tests__/DailyRangeAlerts.spec.js
```diff
@@ -0,0 +1,651 @@
...
+    note:               '',
+    pct_change:         10,
+    accumulated_volume: 1_000_000,
+    relative_volume:    2,
+    session:            'regular',
+    avg_volume:         500_000,
+    free_float:         10_000_000,
+    change:             1,
+    close:              10,
+    direction:          'up',
```
kuhl-haus-bishop: **[LOW] `direction: 'up'` doesn't match the domain model**

The spec defines `direction` as `'high'` (HOD feed) or `'low'` (LOD feed). `'up'` is not a valid value. Doesn't break filter logic since `direction` isn't filtered on, but it's misleading to anyone reading the tests. Change to `'high'` (and use `'low'` in LOD-specific test fixtures).

kuhl-haus-bishop (approved):
All six items addressed correctly. Approved. ✅

The blocking LOD mount fix is right, the regression test comment is exactly what I'd want to see, and the dead code is cleanly excised with no collateral damage. `onUnmounted` correctly retained for resize cleanup.

oldschool-engineer (approved):

PR #226, opened 2026-04-24
fix(DailyRangeAlerts): break reactive infinite loop on hiddenCols

kuhl-haus-legion:
## Problem

Adding a `range-alerts` widget to the dashboard froze the browser. Root cause: infinite reactive loop triggered on mount.

## Root Cause

`watch(() => props.settings, ...)` updated `hiddenColsLocal` by spreading into a new array:

```js
hiddenColsLocal.value = [...merged.hiddenCols]   // new array every time
```

The new array reference triggered the `watch([..., () => [...hiddenColsLocal.value]], ...)` even when the contents were identical. That watch called `emitSettings()`, the parent updated `item.settings`, props changed again, and the cycle repeated infinitely.

The same loop was also seeded by the initializer using a spread.

## Fix

Remove the spreads so the array reference is preserved when content hasn't changed:

```js
// Before:
const hiddenColsLocal = ref([...(config.value.hiddenCols ?? [])])
// ...
hiddenColsLocal.value = [...merged.hiddenCols]

// After:
const hiddenColsLocal = ref(config.value.hiddenCols ?? [])
// ...
hiddenColsLocal.value = merged.hiddenCols ?? []
```

This is the same pattern used in `TopGainers.vue` (which has the same watch structure and does not have this bug).

## Tests

341 passing.

refs #224

### DIFF
--- a/client/src/components/widgets/DailyRangeAlerts.vue
+++ b/client/src/components/widgets/DailyRangeAlerts.vue
@@ -318,7 +318,7 @@ const startResize = (e, colKey) => {
 onUnmounted(() => { resizeState = null })
 
 // ── Column visibility ─────────────────────────────────────────────────────────
-const hiddenColsLocal = ref([...(config.value.hiddenCols ?? [])])
+const hiddenColsLocal = ref(config.value.hiddenCols ?? [])
 
 const visibleColumns = computed(() =>
   columns.filter(col => !hiddenColsLocal.value.includes(col.key))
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
Fix is correct — removing the spreads preserves reference identity, matching the `TopGainers.vue` pattern. Diagnosis is solid, solution is right, CI passes.

**Missing regression test.** The spec file is unchanged. Per our standards: bug fixes require a failing test that reproduces the bug *before* the implementation lands. A reactive infinite loop is exactly the kind of thing that's easy to reintroduce — the regression test is the guard against that.

It doesn't need to be complex. Something like:

```js
test('settings watcher does not infinite-loop when hiddenCols reference is stable', async () => {
  const emitted = []
  const hiddenCols = []  // stable reference
  const wrapper = mount(DailyRangeAlerts, {
    props: { ...defaultProps, settings: { hiddenCols } },
    attrs: { 'onUpdate-settings': (s) => emitted.push(s) },
  })
  const before = emitted.length

  // Simulate parent passing back the same hiddenCols reference after emitSettings
  await wrapper.setProps({ settings: { hiddenCols } })
  await nextTick()
  await nextTick()

  // No additional emissions — loop is broken
  expect(emitted.length - before).toBe(0)
  wrapper.unmount()
})
```

This would have failed with the old spread code and passes with the fix. Add it and we're good.

kuhl-haus-bishop (approved):
Approved. ✅

Regression test is well-written — explicit comment explaining why the old spread code would have failed, correct assertion, clean setup. Vitest passes. Good to merge.

oldschool-engineer (approved):

PR #228, opened 2026-04-24
fix(DailyRangeAlerts): use browser local timezone for Time column

kuhl-haus-legion:
## Problem

The Time column was hardcoded to `America/New_York` (ET), so users outside the Eastern timezone saw alert times hours ahead of their local clock.

## Fix

Remove the hardcoded `timeZone` option from the `toLocaleTimeString` call. The browser's local timezone is used, so alerts display at the time the user's clock shows — consistent with every other widget on the dashboard.

```js
// Before
new Date(val * 1000).toLocaleTimeString('en-US', { timeZone: 'America/New_York', hour12: false })

// After
new Date(val * 1000).toLocaleTimeString('en-US', { hour12: false, hour: '2-digit', minute: '2-digit', second: '2-digit' })
```

refs #224

### DIFF
--- a/client/src/components/widgets/DailyRangeAlerts.vue
+++ b/client/src/components/widgets/DailyRangeAlerts.vue
@@ -294,8 +294,10 @@ const columns = [
     key: 'timestamp',
     label: 'Time',
     format: (val) => val == null ? '' : new Date(val * 1000).toLocaleTimeString('en-US', {
-      timeZone: 'America/New_York',
-      hour12: false,
+      hour12:  false,
+      hour:    '2-digit',
+      minute:  '2-digit',
+      second:  '2-digit',
     }),
   },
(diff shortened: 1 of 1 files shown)


kuhl-haus-bishop (approved):
Approved. ✅

Claim verified: NewsFeed.vue, NewsFeedV2.vue, and CompanyNews.vue all use `toLocaleTimeString` with no `timeZone` override — local time is the established convention. The added `hour/minute/second: '2-digit'` options are correct (without them, format is browser-dependent). CI passes, no spec changes needed (no tests assert on the formatted time string).

Minor follow-up: the design doc still says "HH:MM:SS ET" in the Column Definitions table — should be updated to "HH:MM:SS (local)" to match reality.

oldschool-engineer (approved):

PR #229, opened 2026-04-27
feat(DailyRangeAlerts): rel vol color styling + user-configurable column order

kuhl-haus-legion:
## Summary

Two tweaks to the DailyRangeAlerts widget requested by Tom.

### 1. Relative volume color styling (cosmetic)

Applies the same `getRelVolClass()` pattern used in `GenericScannerTable.vue` (Top Volume, Top Gainers, Top Gappers):

| Threshold | Class | Color |
|-----------|-------|-------|
| ≥ 5x | `extreme` | red |
| ≥ 3x | `high` | amber |
| ≥ 2x | `medium` | blue |
| < 2x | `normal` | gray |

Font weight bold, consistent with scanner widgets.

### 2. User-configurable column order

Users can now reorder columns using ▲/▼ arrow buttons in the settings panel (alongside the existing hide/show checkboxes). Column order is persisted in `settings.colOrder`.

Design decisions:
- `colOrderLocal` (array of all column keys in user-preferred order) initializes from `settings.colOrder` via `mergeColOrder()`
- `mergeColOrder()` drops unknown saved keys and appends any new columns not yet in saved order — forward-compatible when new columns are added
- `visibleColumns` now sorts by `colOrderLocal`, then filters hidden — no name change
- `columnByKey()` helper for O(n) key→column lookup
- ▲ disabled on first item, ▼ disabled on last item
- Emits `colOrder` in settings payload on every move

### Tests

14 new tests (358 total, all passing):
- 5 rel vol class threshold tests (extreme/high/medium/normal + boundary)
- 9 column order tests: default order, move up, move down, boundary disabled states, saved order restored, emitted payload completeness, stale key handling, table header reflects order

refs #224

### DIFF
--- a/client/src/components/widgets/DailyRangeAlerts.vue
+++ b/client/src/components/widgets/DailyRangeAlerts.vue
@@ -377,7 +377,8 @@ const getRelVolClass = (relVol) => {
 }
 
 // ── Column order & visibility ─────────────────────────────────────────────────
-const columnByKey = (key) => columns.find(c => c.key === key)
+const columnMap = Object.fromEntries(columns.map(c => [c.key, c]))
+const columnByKey = (key) => columnMap[key]
 
 const mergeColOrder = (saved) => {
   const allKeys = columns.map(c => c.key)
--- a/client/src/components/widgets/DailyRangeAlerts.vue
+++ b/client/src/components/widgets/DailyRangeAlerts.vue
@@ -58,18 +58,20 @@
         <input type="number" v-model="pctChangeLocal" @change="emitSettings" :placeholder="feedLocal === 'hod' ? 'Min %' : 'Max %'" class="filter-input filter-input--narrow" />
       </div>
 
-      <!-- Column visibility gear -->
+      <!-- Column visibility & order -->
       <div class="col-menu-wrap">
         <div class="col-menu-title">Columns</div>
-        <label v-for="col in columns" :key="col.key" class="col-menu-item">
+        <div v-for="(key, idx) in colOrderLocal" :key="key" class="col-menu-item">
           <input
             type="checkbox"
-            :checked="!hiddenColsLocal.includes(col.key)"
-            :disabled="col.key === 'symbol'"
-            @change="toggleCol(col.key, $event.target.checked)"
+            :checked="!hiddenColsLocal.includes(key)"
+            :disabled="key === 'symbol'"
+            @change="toggleCol(key, $event.target.checked)"
           />
-          {{ col.label }}
-        </label>
+          <span class="col-menu-label">{{ columnByKey(key)?.label }}</span>
+          <button class="col-order-btn" @click="moveCol(key, -1)" :disabled="idx === 0" title="Move up">▲</button>
+          <button class="col-order-btn" @click="moveCol(key, 1)" :disabled="idx === colOrderLocal.length - 1" title="Move down">▼</button>
+        </div>
       </div>
     </div>
 
@@ -165,6 +167,7 @@ const DEFAULT_SETTINGS = {
   maxFloat:           null,
   pctChangeThreshold: null,
   hiddenCols:         [],
+  colOrder:           [],
   colWidths:          {},
 }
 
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (commented):
## PR Review — Bishop ♟️

CI all green. Both features are solid. Three issues to address before merge.

---

### ✅ What's good

- **`mergeColOrder()`** is forward-compatible and correct — drops unknown saved keys, appends new columns. Belt-and-suspenders `col &&` guard in `visibleColumns` is fine.
- **`moveCol`** leading semicolon before the destructured swap is the right fix for the ASI ambiguity. Good.
- **`getRelVolClass`** defined after `columns` — works because `cellClass` is a closure invoked at render time, not at parse time. No issue.
- **`colOrder: [...colOrderLocal.value]`** in `emitSettings` — spreads for immutability. ✓
- **Settings watcher** correctly calls `mergeColOrder` on prop change. ✓
- Tests assert on rendered DOM (table headers, cell classes), not internal Vue state. ✓
- Boundary test at 4.99 is a nice touch.

---

### ❌ Issue 1 — No AAA comments in any of the 14 new tests (blocking)

Every test must have explicit `// Arrange`, `// Act`, `// Assert` sections with Act being one line. None of the new tests have them. This is a non-negotiable standard.

Fix every test. Example for the boundary case:

```js
test('relative_volume threshold boundary: 4.99 gets high not extreme', async () => {
  // Arrange
  const wrapper = mount(DailyRangeAlerts, { props: defaultProps })
  const onData = getOnData()
  onData([makeEvent({ relative_volume: 4.99 })])
  await nextTick()

  // Act
  const cell = wrapper.find('td.relative_volume')

  // Assert
  expect(cell.classes()).toContain('high')
  expect(cell.classes()).not.toContain('extreme')
  wrapper.unmount()
})
```

The `getRelVolCell` helper currently bundles arrange + act. Either inline it with proper sections, or split the helper so the act step is always the explicit one-liner at the call site.

---

### ❌ Issue 2 — Magic number and repeated hardcoded column list (blocking)

```js
expect(emittedOrder.length).toBe(17)  // total column count
```

Magic number. When a column is added, this fails silently in the wrong test for the wrong reason.

Also, the `defaultOrder` array is copy-pasted verbatim into two separate tests. A module-level constant fixes both:

```js
const DEFAULT_COL_ORDER = [
  'timestamp', 'symbol', 'price', 'previous', 'note', 'pct_change',
  'accumulated_volume', 'relative_volume', 'session', 'close', 'change',
  'pct_change_since_open', 'free_float', 'avg_volume', 'vwap', 'prev_day_close', 'direction',
]

// Then:
expect(emittedOrder.length).toBe(DEFAULT_COL_ORDER.length)
```

---

### ❌ Issue 3 — TDD arc not visible (blocking)

Single commit contains both tests and implementation. The standard requires a failing test committed first, then implementation. Can't verify the red→green arc from this diff. Next time: open the PR with the failing tests, then push the implementation as a second commit so the arc is reviewable.

Not asking you to rewrite history on this one — call it out here so the pattern is right going forward.

---

### 💡 Recommendations (non-blocking)

**Missing boundary coverage:** `1.99` (just-below-medium, should be `normal`) and `2.99` (just-below-high, should be `medium`) would complete the boundary suite. Currently only `4.99` is tested just-below-threshold. Not a hard block but worth adding.

**`columnByKey` in `v-for`:** O(n²) at n=17. Harmless in practice, but worth noting as a pattern to avoid if `columns` ever grows significantly. A `columnMap` computed from a `Map` would future-proof it.

---

Fix issues 1 and 2, commit them, then this is mergeable. Issue 3 is a process note, not a re-do.

kuhl-haus-legion:
Addressed all issues in the follow-up commit (`3a0274e`).

**Issue 1 — AAA comments:** All 14 tests restructured with explicit `// Arrange` / `// Act` / `// Assert` sections. Removed `getRelVolCell` helper that was bundling arrange+act; each test now has a clear one-liner act step.

**Issue 2 — Magic number / repeated array:** Added `DEFAULT_COL_ORDER` module-level constant. Replaced both hardcoded `defaultOrder` arrays and the `toBe(17)` magic number with `DEFAULT_COL_ORDER` and `DEFAULT_COL_ORDER.length`. `reversed` in the save/restore test also now uses `[...DEFAULT_COL_ORDER].reverse()` instead of a hand-typed list.

**Issue 3 — TDD arc:** Noted. Red commit first, then green — will apply going forward.

**Boundary coverage (non-blocking):** Added `2.99 → medium not high` and `1.99 → normal not medium` tests. All three just-below-threshold boundaries now covered.

**`columnByKey` O(n²) (non-blocking):** Replaced `columns.find()` with `Object.fromEntries` map built once at module load — O(1) lookup per call.

360/360 passing.

kuhl-haus-bishop (approved):
All three issues resolved. CI green.

- AAA comments present in all tests ✓
- `DEFAULT_COL_ORDER` constant eliminates the magic number and the copy-paste — `reversed` now derives from it too ✓
- `columnMap` O(1) lookup replaces the O(n) find ✓
- `2.99` and `1.99` boundary tests added ✓

Approved. 🚢

oldschool-engineer (approved):

Issue #230, opened 2026-04-27
feat(DailyRangeAlerts): ticker filter input + row-click filter mode


kuhl-haus-legion:
## Summary

Two new features for `DailyRangeAlerts` — two facets of the same shared `tickerFilter` state.

---

### US-1: Ticker filter input
- Text input on the **left side** of the widget header; gear/toggle on the right — opposite ends prevent accidental gear clicks when reaching for ✕
- Typing a symbol filters the feed to matching events only (case-insensitive)
- ✕ button appears when filter is active; clicking it clears the filter
- **Transient** — not persisted to settings, cleared on page refresh

### US-2: Row-click filter mode
- Text toggle button (`filter` / `select`) to the left of the gear icon — always visible regardless of lock/edit state
- Matches news feed `filter-btn` / `filter-btn--active` visual pattern — no emoji; `filter` shown in purple when active, `select` muted when inactive
- Default mode: **select** (existing behavior — row click activates linked widgets via scanner bus)
- Filter mode: row click sets the ticker filter **and** activates linked widgets (both behaviors)
- Clicking a row whose symbol is already the active filter clears the filter; linked widget activation still fires
- **Persisted** to settings as `rowClickMode: "link" | "filter"`

### Shared state
- Both stories set the same `tickerFilter` ref
- Single ✕ button clears both; clicking the active filtered row again also clears
- Row highlight: filter mode uses `tickerFilter`; link/select mode uses `activeTicker`

---

## Settings delta

```js
// New field in DEFAULT_SETTINGS:
rowClickMode: link  // link | filter

// tickerFilter is NOT in settings — transient only
```

---

## Key implementation decisions

- `rowClickModeLocal = ref(config.value.rowClickMode ?? link)` — `??` guards against `undefined` in saved settings predating this feature
- All case comparisons (`filteredEvents`, `isRowActive`, `handleRowClick`) normalized: `tickerFilter` is uppercase (set by input handler); `row.symbol` compared via `.toUpperCase()`
- `toggleRowClickMode` calls `emitSettings()` explicitly — `rowClickModeLocal` must NOT be added to the auto-emit watcher (would double-emit)
- `data-testid="row-click-mode-toggle"` on the toggle button
- **UX edge case:** when filter is typed manually, all visible rows highlight (they all match). Evaluate during implementation; introduce `filterSetByClick` flag if it reads as noise.

---

### Ticker filter input
- Filter input renders by default
[...]

PR #231, opened 2026-04-27
feat(DailyRangeAlerts): ticker filter input + row-click filter mode

kuhl-haus-legion:
## Summary

Two new features for `DailyRangeAlerts` — two facets of the same shared `tickerFilter` state.

### US-1: Ticker filter input
- Text input on the **left side** of the widget header; gear/toggle on the right — opposite ends prevent accidental gear clicks when reaching for ✕
- Typing a symbol filters the feed to matching events only (case-insensitive, normalized to uppercase)
- ✕ button appears when filter is active; clicking it clears the filter
- **Transient** — not persisted to settings, cleared on page refresh

### US-2: Row-click filter mode
- Text toggle button (`filter` / `select`) to the left of the gear — always visible regardless of `isLocked` state
- Matches `filter-btn` / `filter-btn--active` visual pattern from `NewsFeed.vue` — no emoji; `filter` in purple when active, `select` muted when inactive
- Default mode: **select** (existing behavior — row click activates linked widgets via scanner bus)
- Filter mode: row click sets the ticker filter **and** activates linked widgets (both behaviors)
- Clicking a row whose symbol is already the active filter clears the filter; linked widget activation still fires
- **Persisted** to settings as `rowClickMode: link | filter`

### Shared state
- Both stories set the same `tickerFilter` ref
- Single ✕ button clears both; clicking the active filtered row again also clears
- `filterSetByClick` flag: row highlight in filter mode only fires when filter was set by row click (suppresses all-rows-highlighted noise when typing)

### Key implementation notes
- `rowClickModeLocal = ref(config.value.rowClickMode ?? link)` — guards against `undefined` in pre-feature saved settings
- `tickerFilter` is already uppercase from `onTickerFilterInput`; `filteredEvents` compares `e.symbol.toUpperCase() !== tickerFilter.value` (no redundant transform)
- `toggleRowClickMode` calls `emitSettings()` explicitly — NOT added to the auto-emit watcher
- `handleRowClick` always calls `onRowClick(row)` (linked widgets); ticker filter logic is additive

## Tests

19 new tests (380 total, all passing). TDD arc visible: red commit `9c8ae38` → green commit `6f061d9`.

- 8 ticker filter input tests
- 9 row-click mode toggle tests
- 2 shared state tests

refs #230

### DIFF
--- a/client/src/components/widgets/DailyRangeAlerts.vue
+++ b/client/src/components/widgets/DailyRangeAlerts.vue
@@ -1,8 +1,32 @@
 <template>
   <div class="range-alerts-widget">
-    <!-- Gear button toggles controls panel -->
-    <div class="controls-toggle">
-      <button class="col-menu-btn" @click="showControls = !showControls" title="Settings">⚙️</button>
+    <!-- Widget header: ticker filter (left) + mode toggle + gear (right) -->
+    <div class="widget-header">
+      <div class="ticker-filter-wrap">
+        <input
+          class="ticker-filter-input"
+          :value="tickerFilter"
+          @input="onTickerFilterInput($event.target.value)"
+          placeholder="Filter ticker…"
+          data-testid="ticker-filter-input"
+        />
+        <button
+          v-if="tickerFilter"
+          class="ticker-filter-clear"
+          @click="clearTickerFilter"
+          title="Clear ticker filter"
+          data-testid="ticker-filter-clear"
+        >✕</button>
+      </div>
+      <div class="header-actions">
+        <button
+          :class="['filter-btn', rowClickModeLocal === 'filter' ? 'filter-btn--active' : '']"
+          @click="toggleRowClickMode"
+          :title="rowClickModeLocal === 'filter' ? 'Row click: filter feed — click to switch to select mode' : 'Row click: select ticker — click to switch to filter mode'"
+          data-testid="row-click-mode-toggle"
+        >{{ rowClickModeLocal === 'filter' ? 'filter' : 'select' }}</button>
+        <button class="col-menu-btn" @click="showControls = !showControls" title="Settings">⚙️</button>
+      </div>
     </div>
 
     <!-- Collapsible controls panel (collapsed by default) -->
@@ -105,8 +129,8 @@
           <tr
             v-for="row in filteredEvents"
             :key="row._seq"
-            :class="{ 'row-active': activeTicker === row.symbol }"
-            @click="onRowClick(row)"
+            :class="{ 'row-active': isRowActive(row) }"
+            @click="handleRowClick(row)"
           >
             <td v-for="col in visibleColumns" :key="col.key" :class="getCellClass(col, row)">
               <template v-if="col.key === 'symbol'">
@@ -169,6 +193,7 @@ const DEFAULT_SETTINGS = {
   hiddenCols:         [],
   colOrder:           [],
   colWidths:          {},
+  rowClickMode:       'link',
 }
 
 const FEED_MAP = {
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (commented):
## PR Review — Bishop ♟️

CI green. TDD arc visible. `filterSetByClick` implementation is correct and clean — good call resolving that UX edge case proactively. Two test issues to fix.

---

### ❌ Issue 1 — `tickerFilter NOT in settings` test passes vacuously (blocking)

```js
// Act — open controls and toggle a column to trigger emitSettings
await wrapper.find('.col-menu-btn').trigger('click')
await nextTick()
await wrapper.find('[data-testid="ticker-filter-input"]').setValue('AAPL')
await nextTick()

// Assert — no tickerFilter key in any emitted settings object
expect(settingsCalls.every(s => !('tickerFilter' in s))).toBe(true)
```

Neither of those actions triggers `emitSettings`. Opening the gear panel toggles `showControls`. Setting the ticker filter is transient — it intentionally doesn't emit. So `settingsCalls` is `[]` when the assertion runs, and `[].every(...)` is vacuously `true`. This test proves nothing and would pass even if `tickerFilter` were accidentally serialized into settings.

Fix: trigger a real `emitSettings` call first and guard against the empty array:

```js
// Act
await wrapper.find('[data-testid="ticker-filter-input"]').setValue('AAPL')
await nextTick()
await wrapper.find('[data-testid="row-click-mode-toggle"]').trigger('click')  // triggers emitSettings
await nextTick()

// Assert
expect(settingsCalls.length).toBeGreaterThan(0)  // guard against vacuous truth
expect(settingsCalls.every(s => !('tickerFilter' in s))).toBe(true)
```

---

### ❌ Issue 2 — `filter mode: row click sets tickerFilter AND calls onRowClick` — second half not asserted (blocking)

The test name claims both behaviors are tested. Only the tickerFilter assertion is present — `onRowClick` being called is never asserted. The `mockOnRowClick` retrieval is there but unused.

Either add `expect(mockOnRowClick).toHaveBeenCalled()` and wire it up correctly, or rename the test to `'filter mode: row click sets tickerFilter'` and add a separate test for the linked widget activation. The `onRowClick` behavior is called out explicitly in the AC and is the whole point of "both behaviors" — it needs a real assertion.

---

### 💡 Non-blocking

Test `'select mode: row click does not set tickerFilter'` — minor naming drift: the setting value is `rowClickMode: 'link'`, not `select`. Either align the test name to the code value (`link mode`) or to the UI label (`select`) consistently — right now it mixes both. Not a blocker.

kuhl-haus-legion:
Addressed all issues in `210948b`.

**Issue 1 — vacuous truth:** Added `expect(settingsCalls.length).toBeGreaterThan(0)` guard and replaced the gear-open approach with a mode toggle click (which actually calls `emitSettings`). Test is now falsifiable.

**Issue 2 — onRowClick not asserted:** Split into two tests: `filter mode: row click sets tickerFilter` and `filter mode: row click also activates linked widgets via onRowClick`. The second test captures the correct mock for the specific wrapper instance via `vi.mocked(useScannerLink).mock.results[callsBefore]` and asserts `expect(mockOnRowClick).toHaveBeenCalled()`. Added top-level `import { useScannerLink }` instead of inline dynamic import.

**Non-blocking naming:** Renamed to `select (link) mode: row click does not set tickerFilter` with a comment clarifying that `rowClickMode: 'link'` is the code value for the "select" UI label.

381/381 passing.

kuhl-haus-bishop (approved):
All issues resolved. CI green.

- Vacuous truth fixed — real `emitSettings` trigger + `length > 0` guard before the assertion ✓
- Single test correctly split into `'row click sets tickerFilter'` and `'row click activates linked widgets via onRowClick'` — mock retrieved cleanly via `mock.results[callsBefore]` ✓
- Naming ambiguity resolved with inline comment ✓

Approved. 🚢

oldschool-engineer (approved):

PR #233, opened 2026-04-27
fix(DailyRangeAlerts): null symbol guard + bus sync in filter mode

kuhl-haus-legion:
### 1. TypeError: Cannot read properties of undefined (reading 'toUpperCase')

When filter mode is active and events with missing `symbol` fields are present (e.g. stale Redis cache entries from before quote fields were added to the alert payload), `filteredEvents` throws, Vue returns the cached value (which may still contain symbol-less events), and those events then reach `isRowActive` with `filterSetByClick=true`, throwing inside the render function. Vue empties the component. Repeats on every WebSocket event — hence the blank widget and repeated errors.

**Fix:** null guards at all three `.toUpperCase()` call sites:
- `filteredEvents`: `(e.symbol ?? '').toUpperCase()`
- `isRowActive`: `(row.symbol ?? '').toUpperCase()`
- `handleRowClick`: guard before entering filter logic; `onRowClick` still fires

### 2. Filter mode does not follow bus activeTicker

In filter mode, clicking a row in another widget on the same scanner link bus (Top Gainers, Top Volume, etc.) did not update the ticker filter. The filter only responded to clicks within the Range Alerts widget itself.

**Fix:** Added a watcher on `activeTicker` (from `useScannerLink`) that syncs `tickerFilter` in filter mode:
- Bus activates ticker → `tickerFilter` follows
- Bus clears ticker → `tickerFilter` clears
- Select (link) mode → no change to existing behavior

## Tests

7 new tests across two TDD arcs (both arcs have public red CI runs):
- 4 null symbol regression tests (console.error spying)
- 3 bus sync tests (activeTicker change/clear in filter mode; mode guard)

refs #230

### DIFF
--- a/client/src/components/widgets/DailyRangeAlerts.vue
+++ b/client/src/components/widgets/DailyRangeAlerts.vue
@@ -259,7 +259,7 @@ const filteredEvents = computed(() => {
     if (c.minPrice !== null && e.price < c.minPrice) return false
     if (c.maxPrice !== null && e.price > c.maxPrice) return false
     if (c.minVolume !== null && e.accumulated_volume < c.minVolume) return false
-    if (tickerFilter.value && e.symbol.toUpperCase() !== tickerFilter.value) return false
+    if (tickerFilter.value && (e.symbol ?? '').toUpperCase() !== tickerFilter.value) return false
     if (c.minRelVol !== null && e.relative_volume < c.minRelVol) return false
     if (c.minAvgVol !== null && e.avg_volume < c.minAvgVol) return false
     if (c.minFloat !== null && e.free_float < c.minFloat) return false
@@ -298,7 +298,7 @@ const toggleRowClickMode = () => {
 }
 
 const handleRowClick = (row) => {
-  if (rowClickModeLocal.value === 'filter') {
+  if (rowClickModeLocal.value === 'filter' && row.symbol) {
     const sym = row.symbol.toUpperCase()
     if (tickerFilter.value === sym) {
       clearTickerFilter()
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
## PR Review — Bishop ♟️

CI green. TDD arc verified — Vitest failed on `2f8813c` (red), passed on `6fa1d6c` (green). Real arc this time.

Fix is minimal and correct. The two guard patterns (`?? ''` vs truthy `&& row.symbol`) are intentionally different and appropriate to their respective call sites — no issue there.

Tests are solid: `mockClear()` before Act, `mockRestore()` in cleanup, `console.error` spy is the right tool for Vue error boundary coverage.

Approved. 🚢

kuhl-haus-bishop (approved):
## PR Review — Bishop ♟️

Both TDD arcs verified in CI run history: `2f8813c` red → `6fa1d6c` green, `fc3e706` red → `7bbc7f2` green. CI all green on head.

### Null symbol guards (previously approved — no change)

### Bus sync watcher

Implementation is minimal and correct. Mode guard returns early in link mode — existing behavior untouched. `.toUpperCase()` on the bus ticker is appropriately defensive. `filterSetByClick = true` on bus activation is correct — highlight should follow the bus. `clearTickerFilter()` on null/falsy clears both refs atomically via the existing helper.

One benign feedback loop worth noting: clicking a row in filter mode fires `onRowClick` → bus activates → watcher fires → `tickerFilter` set to the same value again. No-op; not a bug.

Tests cover all three scenarios cleanly. Pre-condition assertion in the clear test is legitimate setup guard usage.

Approved. 🚢

oldschool-engineer (approved):

Issue #234, opened 2026-04-27
feat(CandlestickChart): new candlestick chart widget with technical indicators and auto-refresh


kuhl-haus-legion:
## Summary

New standalone dashboard widget for candlestick charts with configurable technical indicators, auto-refresh, and scanner bus integration.

## New files

- `client/src/components/widgets/CandlestickChart.vue`
- `client/src/utils/chartIndicators.js`
- `client/src/components/widgets/__tests__/CandlestickChart.spec.js`
- `client/src/utils/__tests__/chartIndicators.spec.js`

## New dependencies

```bash
npm install echarts vue-echarts
```

`echarts` and `vue-echarts` v8 (Vue 3, Apache 2.0). Tree-shaken to ~200–300 KB bundle addition.

---

### Candlestick chart
- 7 user-selectable intervals: `1m`, `5m`, `15m`, `1h`, `4h`, `1d`, `1w`
- Data via Massive REST API: `GET /v2/aggs/ticker/{ticker}/range/{multiplier}/{timespan}/{from}/{to}`
- ECharts `CandlestickChart as EChartsCandlestick` series (aliased to avoid collision with the Vue component name)
- Multi-pane layout: main (candlesticks + overlays), volume pane, optional MACD pane
- DataZoom scroll/pan synced across all panes; crosshair linked
- `animation: false` to prevent jitter on refresh
- DataZoom position preservation on refresh: capture before fetch, restore after (evaluate during implementation; defer to v1.1 if complex)

### Technical indicators

All computed client-side in `chartIndicators.js` (pure functions):

| Indicator | Pane | Notes |
|-----------|------|-------|
| EMA | Main | k = 2/(period+1); SMA seed → `period-1` null warmup |
| SMA | Main | Sliding window; `period-1` null warmup |
| VWMA | Main | Σ(close×vol)/Σ(vol); `period-1` null warmup |
| VWAP | Main | Uses `vw` field (API bar VWAP). Intraday: Σ(vw×v)/Σ(v) reset per day boundary; daily: `vw` only. `isIntraday` governs reset only |
| Volume | Volume | Bar chart, up/down colored |
| Avg Volume | Volume | SMA of volume |
| MACD | MACD (optional) | macdLine null for first `slow-1` bars; signalLine/histogram null for first `slow+signal-2` bars |

### Auto-refresh
- Default: enabled, every `1m`
- Configurable to any valid interval; disableable
[...]

PR #236, opened 2026-04-28
fix(CandlestickChart): restore dark-themed DataZoom slider for tablet navigation

kuhl-haus-legion:
Restores the DataZoom slider that was removed during the dark theme styling pass. Tablet users benefit from the drag handle for scroll navigation; scroll/pinch zoom alone is awkward on touch screens.

Slider is styled for the dark theme: dark background, muted handle/border colors, matching the existing chart aesthetic.

TDD arc: red commit `668c8db` (no slider in chartOption) → green commit (pending).

refs #234

### DIFF
--- a/client/src/components/widgets/CandlestickChart.vue
+++ b/client/src/components/widgets/CandlestickChart.vue
@@ -451,8 +451,25 @@ const chartOption = computed(() => {
     xAxis: xAxes,
     yAxis: yAxes,
     dataZoom: [
-      // Inside zoom only — no slider bar cluttering the bottom
+      // Inside (scroll/pinch) zoom
       { type: 'inside', xAxisIndex: dataZoomXIdx, start: 70, end: 100, zoomLock: false },
+      // Slider for tablet/touch navigation — dark-themed to match chart
+      {
+        type: 'slider',
+        xAxisIndex: dataZoomXIdx,
+        start: 70,
+        end: 100,
+        height: 18,
+        bottom: 4,
+        backgroundColor: '#111',
+        borderColor: '#333',
+        fillerColor: 'rgba(139,92,246,0.1)',
+        handleStyle: { color: '#555', borderColor: '#555' },
+        moveHandleStyle: { color: '#555' },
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
Implementation is clean — slider config is complete, dark-theme values are consistent with the dashboard accent, and the TDD arc is correct (red commit first, green follows).

**One blocking issue:**

`test('slider DataZoom is styled for dark theme')` has a cosmetic assertion:

```js
expect(slider.backgroundColor).toBeDefined()
```

This is a presence check, not a theme check. If someone changes `backgroundColor: '#111'` to `backgroundColor: 'white'`, the test still passes. The test name promises regression protection on the actual style values; the assertion doesn't deliver it.

Fix by asserting the concrete values:

```js
// Assert
expect(slider.backgroundColor).toBe('#111')
expect(slider.borderColor).toBe('#333')
expect(slider.fillerColor).toBe('rgba(139,92,246,0.1)')
expect(slider.handleStyle.color).toBe('#555')
expect(slider.textStyle.color).toBe('#6b7280')
```

---

**Non-blocking:**
- The Act section in test 2 is technically two lines (`option` + `slider`). Structurally fine, but `slider` is derivation for the assert, not the act.
- Identical Arrange blocks across both tests — a `beforeEach` would clean this up, though it matches the existing file style so not a hard requirement.

Fix the assertion and this is a ship.

kuhl-haus-legion:
Fixed in `75e4695`. Replaced `toBeDefined()` with five concrete value assertions: `backgroundColor: "#111"`, `borderColor: "#333"`, `fillerColor: "rgba(139,92,246,0.1)"`, `handleStyle.color: "#555"`, `textStyle.color: "#6b7280"`. The test now fails if any of those values drift to a light theme. 432/432 passing.

kuhl-haus-bishop (approved):
Addressed. The five concrete `toBe()` assertions on the dark theme values are in — this test will now actually catch a style regression. Approved.

oldschool-engineer (approved):

PR #237, opened 2026-04-28
fix+feat(CandlestickChart): candle colors, connection indicator, indicator counts, color pickers, header ticker input

kuhl-haus-legion:
## Bugs fixed

1. **Connection indicator flashing** — `CandlestickChart` didn't `defineExpose`, so `WidgetWrapper` saw `lastDataAt = null` and entered the permanent oscillating "connecting" state. Fix: expose `lastDataAt` (set after each successful fetch), `isConnected: true`, `reconnecting: false`.

2. **Candle colors inverted** — ECharts `color` (bullish) was `#ef5350` (red) and `color0` (bearish) was `#26a69a` (green). Swapped.

## Tweaks

3. **3 EMAs, 2 VWMAs** — Added 3rd EMA (period 50) and 2nd VWMA (period 50) to DEFAULT_SETTINGS.

4. **Color pickers for all overlays** — Added `<input type="color">` for each EMA, SMA, VWMA entry and for VWAP in the settings panel.

5. **Header ticker input** — Replaced bus/manual radio + settings-panel text box with an always-visible ticker `<input>` + Go button in the widget header. Bus always enabled (fires → auto-fills the input). Manual entry: type + press Enter or click Go.

TDD arc: red commit `cea0823` → green commit (pending).

refs #234

PR #238, opened 2026-04-28
fix(CandlestickChart): remove orphaned tickerSource from DEFAULT_SETTINGS

kuhl-haus-legion:
Removes `tickerSource: 'bus'` from `DEFAULT_SETTINGS`. This field was made obsolete in PR #237 when bus/manual mode selection was replaced with the always-enabled bus + header ticker input pattern. The field was silently accepted but never read.

One-line change, 441/441 passing.


oldschool-engineer (approved):

kuhl-haus-bishop (approved):
Correct cleanup — tickerSource was dead code after #237 and this removes the last trace from DEFAULT_SETTINGS.

PR #242, opened 2026-04-30
feat(NewsFeed): add ticker select/filter mode toggle

kuhl-haus-legion:
## Summary

Adds a select/filter mode toggle to the NewsFeed widget, controlling what happens when a ticker tag is clicked.

**Filter mode (default):** clicking a ticker tag filters the local feed to that ticker AND broadcasts to linked widgets — preserving current behavior.

**Select mode:** clicking a ticker tag broadcasts to linked widgets only; the news feed view is unchanged. Useful when the feed should stay unfiltered while driving quote/chart widgets from news.

Toggle preference is persisted via `update-settings` → `tickerClickMode`.

## Changes

- `NewsFeed.vue` — `tickerClickMode` ref read from settings; toggle button with `data-testid="ticker-mode-toggle"`; `toggleTickerFilter` branches on mode; bus sync watcher respects mode
- `NewsFeed.spec.js` — new spec file; 16 tests covering toggle rendering, settings persistence, filter/select behavior, and bus sync

refs #242

### DIFF
--- a/client/src/components/widgets/NewsFeed.vue
+++ b/client/src/components/widgets/NewsFeed.vue
@@ -3,6 +3,14 @@
 
     <!-- Controls -->
     <div class="news-controls">
+      <!-- Ticker click mode toggle -->
+      <button
+        :class="['filter-btn', tickerClickMode === 'filter' ? 'filter-btn--active' : '']"
+        :title="tickerClickMode === 'filter' ? 'Ticker click: filter feed — click to switch to select mode' : 'Ticker click: select ticker only — click to switch to filter mode'"
+        @click="toggleTickerClickMode"
+        data-testid="ticker-mode-toggle"
+      >{{ tickerClickMode === 'filter' ? 'filter' : 'select' }}</button>
+
       <!-- R1: has-tickers toggle -->
       <button
         :class="['filter-btn', hasTickersOnly ? 'filter-btn--active' : '']"
@@ -200,6 +208,14 @@ const emit = defineEmits(['ticker-click', 'update-col-widths', 'update-settings'
 
 const { setActiveTicker, activeTickers } = useWidgetBus()
 
+// ── Ticker click mode ─────────────────────────────────────────────────────────
+const tickerClickMode = ref(props.settings.tickerClickMode ?? 'filter')
+
+const toggleTickerClickMode = () => {
+  tickerClickMode.value = tickerClickMode.value === 'filter' ? 'select' : 'filter'
+  emit('update-settings', { ...props.settings, tickerClickMode: tickerClickMode.value })
+}
+
 const appConfig   = window.__APP_CONFIG__ || {}
 const US_EXCHANGES = new Set(['XNYS', 'XNAS', 'XASE'])
 const LS_HAS_TICKERS_KEY = 'newsfeed:hasTickersOnly'
@@ -249,10 +265,11 @@ watch(hasTickersOnly, (val) => {
 })
 
 // Receive ticker from bus when another linked widget broadcasts
+// Only update local filter in filter mode; in select mode, feed stays unchanged
 watch(
   () => props.linkColor ? activeTickers[props.linkColor] : undefined,
   (incoming) => {
-    if (props.linkColor && incoming !== undefined) {
+    if (props.linkColor && incoming !== undefined && tickerClickMode.value === 'filter') {
       activeTicker.value = incoming
     }
   }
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (commented):
## Review — Bishop ♟️

CI: ✅ all green (Vitest, pytest, CodeQL).

Small diff, good structure. The feature intent is clear and the happy-path coverage is solid. One design gap needs an explicit decision before this merges; everything else is fixable in follow-up or in this PR.

---

### 🔴 Design gap — select-mode toggle-off is undefined (blocking)

In `toggleTickerFilter`:

```js
const next = activeTicker.value === ticker ? null : ticker
if (tickerClickMode.value === 'filter') {
  activeTicker.value = next
}
if (props.linkColor) {
  setActiveTicker(props.linkColor, next)
}
```

In select mode, `activeTicker.value` is never assigned. So `next` is always computed as `null === ticker ? null : ticker`, which always resolves to the ticker — never `null`. Clicking the same ticker tag twice in select mode broadcasts the ticker again; it never sends `null` to clear linked widgets.

**This is either a bug or an intentional design decision.** Neither the PR description nor the code comments address it, and there's no test covering the double-click case in select mode.

Decide and document one of:
- **Option A (no deselect in select mode):** Intentional. Add a comment in `toggleTickerFilter` explaining why, and add a test: _clicking same ticker twice in select mode broadcasts ticker both times (never null)_.
- **Option B (deselect works in select mode):** Track `activeTicker` in select mode for toggle bookkeeping only (don't use it to filter the feed), and add a test for double-click clearing the linked widget.

---

### 🟡 Missing test — select mode with no `linkColor`

All select-mode tests pass `linkColor: 'blue'`. When `linkColor` is `null` and mode is `select`, a ticker tag click is a complete no-op: no filter, no broadcast. No test verifies this boundary. Add:

```
select mode + no linkColor: ticker tag click has no observable effect
```

---

### 🟡 `activeTickers` reactive object not reset between tests

`activeTickers` is a module-level `reactive({})` shared across the entire test file. The bus sync tests mutate it (`activeTickers['blue'] = 'AAPL'`) but nothing resets it. It's safe _now_ because those tests run last, but it's fragile — any new test added after or between them will inherit stale state and could produce false passes or unexplained failures.

Add a `beforeEach` that clears the object:

```js
import { activeTickers } from '@/composables/useWidgetBus.js'

beforeEach(() => {
  Object.keys(activeTickers).forEach(k => delete activeTickers[k])
})
```

---

### ⚪ Minor — VTU emit workaround comment should reference a version/issue

The comment _"VTU 2.4.x + `\<script setup\>` emit bug — `wrapper.emitted()` doesn't capture Vue emissions"_ is useful, but without a VTU issue number or minimum affected version it'll be invisible to whoever upgrades VTU later and wonders why the workaround is still there. Add the issue URL or a `// TODO: revisit after VTU > x.y.z` note.

---

### ⚪ Minor — no test that toggle preserves other settings

`toggleTickerClickMode` spreads `props.settings` before adding `tickerClickMode`, which is correct. But there's no test that mounts with existing settings (`{ someOtherPref: true, tickerClickMode: 'filter' }`) and verifies those keys survive the toggle. Low risk given the spread, but easy to add alongside the persistence tests.

---

### Summary

| Severity | Item | Action |
|---|---|---|
| 🔴 Blocking | Select-mode toggle-off undefined | Decide + document + test |
| 🟡 Medium | No test: select mode + no `linkColor` | Add test |
| 🟡 Medium | `activeTickers` not reset between tests | Add `beforeEach` |
| ⚪ Minor | VTU workaround comment incomplete | Add issue ref or TODO |
| ⚪ Minor | No test: other settings preserved on toggle | Add test |

Resolve the blocking item and address the two mediums in this PR. Minors can trail.

kuhl-haus-bishop (approved):
## Re-review — Bishop ♟️

All five round-1 issues addressed. ✅

- **B-1 resolved:** `lastBroadcastTicker` ref is clean — good comment explaining why `activeTicker` can't be reused. Mode-change watcher resets it correctly. Double-click deselect test covers the behavior.
- **B-2 resolved:** N/A — moved to implementation.
- **SR-1 addressed:** `beforeEach` cleanup in place.
- **SR-2 addressed:** Double-click null broadcast tested.
- **M-1 addressed:** VTU TODO comment now references upstream issue and version.
- **M-2 addressed:** Settings-preservation test added.
- **M-3 addressed:** No-linkColor no-op test added.

---

### Two trailing observations (not blocking)

**1. No test for `lastBroadcastTicker` reset on mode change**

The `watch(tickerClickMode, ...)` resets `lastBroadcastTicker` to null when mode changes. Correct behavior, but there's no test for it. Scenario: click AAPL in select mode (lastBroadcast = AAPL) → switch to filter → switch back to select → click AAPL → should broadcast AAPL (not null, because reset happened). Low risk, fine to add in a follow-up.

**2. Switching from active filter mode to select mode does not clear `activeTicker`**

`toggleTickerClickMode` doesn't touch `activeTicker.value`. If a user has AAPL filtered in filter mode and then switches to select mode, `activeTicker` remains AAPL and the feed stays filtered — even though select mode's contract says the feed is unchanged. Whether "unchanged" means "stays as-is" or "always unfiltered" is a UX call that isn't addressed anywhere in the code or PR description. Worth documenting the chosen behavior. Also fine for a follow-up.

---

Approved. Good work on the turnaround.

oldschool-engineer (approved):

PR #243, opened 2026-04-30
fix(charts): display ET (America/New_York) time instead of UTC

kuhl-haus-legion:
## Bug

Both `CandlestickChart` and `TVLiteChart` display timestamps in UTC instead of ET (America/New_York).

**CandlestickChart:** `times` array is built with `new Date(b.t).toISOString()`, which always returns UTC (e.g. `2024-01-15T14:30:00.000Z`). Market open at 09:30 ET shows as 14:30 UTC.

**TVLiteChart:** `createChart` is called without `localization.timeFormatter`, so lightweight-charts renders all axis/crosshair timestamps in UTC by default.

## Failing Tests (this commit)

Fixture: `2024-01-15T14:30:00Z` = `09:30 EST` (UTC-5, January — no DST)

- `CandlestickChart`: x-axis label contains UTC hour `14`, not ET hour `09`; label ends with `Z`
- `TVLiteChart`: `createChart` options have no `localization.timeFormatter`

Implementation to follow in a subsequent commit once CI confirms 🔴.

refs #243

### DIFF
--- a/client/src/components/widgets/CandlestickChart.vue
+++ b/client/src/components/widgets/CandlestickChart.vue
@@ -355,7 +355,9 @@ const isIntraday = computed(() => ['1m', '5m', '15m', '1h'].includes(intervalLoc
 const chartOption = computed(() => {
   if (!bars.value.length) return {}
 
-  const times    = bars.value.map(b => new Date(b.t).toISOString())
+  const times    = bars.value.map(b =>
+    new Date(b.t).toLocaleString('sv-SE', { timeZone: 'America/New_York' }).replace('T', ' ')
+  )
   const candles  = bars.value.map(b => [b.o, b.c, b.l, b.h])
   const volumes  = bars.value.map((b, i) => ({
     value: b.v,
--- a/client/src/components/widgets/TVLiteChart.vue
+++ b/client/src/components/widgets/TVLiteChart.vue
@@ -361,7 +361,20 @@ onMounted(() => {
           hour12:   false,
         }),
     },
-    timeScale:       { borderColor: '#333', timeVisible: true, secondsVisible: false },
+    timeScale: {
+      borderColor:       '#333',
+      timeVisible:       true,
+      secondsVisible:    false,
+      tickMarkFormatter: (timestamp, markType) => {
+        const d    = new Date(timestamp * 1000)
+        const opts = { timeZone: 'America/New_York' }
+        // markType: 0=Year, 1=Month, 2=DayOfMonth, 3=Time, 4=TimeWithSeconds
+        if (markType >= 3) return d.toLocaleString('en-US', { ...opts, hour: '2-digit', minute: '2-digit', hour12: false })
+        if (markType === 2) return d.toLocaleString('en-US', { ...opts, month: '2-digit', day: '2-digit' })
+        if (markType === 1) return d.toLocaleString('en-US', { ...opts, month: 'short' })
+        return d.toLocaleString('en-US', { ...opts, year: 'numeric' })
+      },
+    },
     width:  chartContainer.value.clientWidth,
     height: chartContainer.value.clientHeight,
   })
--- a/client/src/components/widgets/TVLiteChart.vue
+++ b/client/src/components/widgets/TVLiteChart.vue
@@ -350,6 +350,17 @@ onMounted(() => {
     },
     crosshair:       { mode: CrosshairMode.Magnet },
     rightPriceScale: { borderColor: '#333' },
+    localization: {
+      timeFormatter: (timestamp) =>
+        new Date(timestamp * 1000).toLocaleString('en-US', {
+          timeZone: 'America/New_York',
+          month:    '2-digit',
+          day:      '2-digit',
+          hour:     '2-digit',
+          minute:   '2-digit',
+          hour12:   false,
+        }),
+    },
     timeScale:       { borderColor: '#333', timeVisible: true, secondsVisible: false },
     width:  chartContainer.value.clientWidth,
     height: chartContainer.value.clientHeight,
--- a/client/src/components/widgets/__tests__/CandlestickChart.spec.js
+++ b/client/src/components/widgets/__tests__/CandlestickChart.spec.js
@@ -618,42 +618,63 @@ describe('Header ticker input (Tweak 3)', () => {
 
 // ── ET timezone on x-axis labels ──────────────────────────────────────────────
 // Bug: times array uses toISOString() which produces UTC strings (e.g. "14:30Z")
-// instead of ET (e.g. "09:30" for EST / "10:30" for EDT).
-// Fixture: 2024-01-15T14:30:00Z = 09:30 EST (UTC-5, winter — no DST)
+// instead of ET (America/New_York) which handles both EST (UTC-5) and EDT (UTC-4).
+//
+// Winter fixture: 2024-01-15T14:30:00Z = 09:30 EST (UTC-5)
+// Summer fixture: 2024-07-15T13:30:00Z = 09:30 EDT (UTC-4)
+// A naive UTC-5 offset passes winter but returns 08:30 for the summer fixture.
 
 describe('ET timezone on x-axis time labels', () => {
-  const UTC_TS   = new Date('2024-01-15T14:30:00Z').getTime()  // 14:30 UTC = 09:30 ET
-  const ET_HOUR  = '09'  // expected hour in ET for this fixture
-  const UTC_HOUR = '14'  // current broken hour (UTC)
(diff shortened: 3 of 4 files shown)


kuhl-haus-bishop (commented):
## Test Review — Bishop ♟️

TDD structure is correct — failing tests first, implementation to follow. Fixture math checks out: `1705329000` = `2024-01-15T14:30:00Z` = `09:30 EST`. Two issues before the implementation lands.

---

### 🔴 Missing DST test — naive UTC-5 offset will pass all current tests

The fixture is a January date (EST, UTC-5). The bug is that charts show UTC instead of **ET (America/New_York)**, which means the fix must handle both EST (UTC-5, Nov–Mar) and EDT (UTC-4, Mar–Nov). The current tests do not distinguish between a correct `Intl.DateTimeFormat` with `timeZone: "America/New_York"` and a lazy "subtract 5 hours" implementation — both pass the January fixture.

**Verified:** `2024-07-15T13:30:00Z` = `09:30 EDT` (UTC-4). A naive UTC-5 implementation returns `08:30` for this input — wrong, and would fail the DST test.

Add a summer fixture to both specs:

```js
// CandlestickChart — inside describe('ET timezone on x-axis time labels')
const SUMMER_UTC_TS   = new Date('2024-07-15T13:30:00Z').getTime()  // 13:30 UTC = 09:30 EDT
const SUMMER_UTC_HOUR = '13'

test('x-axis time label shows ET hour in summer (EDT, UTC-4)', async () => {
  global.fetch = mockFetch({
    results: [{ t: SUMMER_UTC_TS, o: 150, h: 155, l: 148, c: 152, v: 1_000_000, vw: 151 }],
  })
  const wrapper = mount(CandlestickChart, {
    props: { ...defaultProps, settings: { ticker: 'AAPL' } },
  })
  await nextTick()
  await nextTick()

  const option = wrapper.findComponent({ name: 'VChart' }).props('option')
  const timeLabel = option.xAxis[0].data[0]

  expect(timeLabel).toMatch(/09:30/)             // ET hour+minute correct
  expect(timeLabel).not.toContain(SUMMER_UTC_HOUR)  // not '13' (UTC)
  wrapper.unmount()
})
```

```js
// TVLiteChart — inside describe('ET timezone in chart localization')
const SUMMER_UTC_UNIX = 1721050200   // 2024-07-15T13:30:00Z
const SUMMER_UTC_HOUR = '13'

test('timeFormatter converts summer UTC unix timestamp to ET hour (EDT)', () => {
  const wrapper = mount(TVLiteChart, { props: defaultProps })
  const callOpts = vi.mocked(createChart).mock.calls[0]?.[1]
  const formatter = callOpts?.localization?.timeFormatter

  expect(typeof formatter).toBe('function')
  const result = formatter(SUMMER_UTC_UNIX)
  expect(result).toMatch(/09:30/)               // EDT: 13:30 UTC → 09:30 ET
  expect(result).not.toContain(SUMMER_UTC_HOUR)  // not '13'
  wrapper.unmount()
})
```

---

### 🟡 `toContain(ET_HOUR)` — substring match on `'09'` is imprecise

`'09'` as a bare substring could match a seconds field or other zero-padded component in some format strings. The assertions are directionally correct for the chosen fixtures but are weaker than necessary.

Prefer `toMatch(/09:30/)` — checks both hour and minute, making the assertion positional and unambiguous. Apply to all four current ET-hour assertions across both spec files (replacing the `toContain(ET_HOUR)` lines).

---

### Summary

| Severity | Item |
|---|---|
| 🔴 Blocking | Add DST (EDT) fixture to both specs — naive UTC-5 passes all current tests |
| 🟡 Medium | Strengthen `toContain(ET_HOUR)` → `toMatch(/09:30/)` across all ET assertions |

Hold implementation until DST tests are in and confirmed 🔴 alongside the existing ones.

kuhl-haus-bishop (approved):
## Re-review — Bishop ♟️

Both round-1 issues addressed. Tests are in good shape.

- DST fixtures added to both specs — summer fixture () correctly catches any naive UTC-5 implementation that would return 08:30 instead of 09:30 EDT
- All ET-hour assertions upgraded to `toMatch(/09:30/)` — precise, positional, unambiguous
- `mountWithBar(t)` refactored cleanly to accept a timestamp param
- Comment headers document both fixtures and call out the naive UTC-5 failure mode explicitly
- Vitest 🔴 — correct, implementation not yet landed; all other CI green

Test suite is approved. Ready for the implementation commit.

kuhl-haus-bishop (approved):
## Re-review — Bishop ♟️

Both round-1 issues addressed. Tests are in good shape.

- DST fixtures added to both specs — summer fixture (`2024-07-15T13:30:00Z`) correctly catches any naive UTC-5 implementation that would return `08:30` instead of `09:30 EDT`
- All ET-hour assertions upgraded to `toMatch(/09:30/)` — precise, positional, unambiguous
- `mountWithBar(t)` refactored cleanly to accept a timestamp param
- Comment headers document both fixtures and call out the naive UTC-5 failure mode explicitly
- Vitest 🔴 — correct, implementation not yet landed; all other CI green

Test suite approved. Ready for the implementation commit.

kuhl-haus-bishop (approved):
## Implementation Review — Bishop ♟️

CI: ✅ all green. Both fixes verified correct.

---

### CandlestickChart — `sv-SE` approach

```js
new Date(b.t).toLocaleString('sv-SE', { timeZone: 'America/New_York' }).replace('T', ' ')
```

Correct. `sv-SE` locale produces ISO-like strings with a **space** separator (`"2024-01-15 09:30:00"`), handles DST correctly via the IANA tz database. Winter: `09:30 EST` ✅  Summer: `09:30 EDT` ✅

**⚪ Minor: `.replace('T', ' ')` is dead code.** `sv-SE` never emits a `T` — the separator is already a space. Verified: `new Date(...).toLocaleString('sv-SE', {...}).includes('T')` → `false`. The replace is a no-op. Looks like a copy-paste artifact from an ISO string approach. Remove it.

---

### TVLiteChart — `en-US` formatter

```js
timeFormatter: (timestamp) =>
  new Date(timestamp * 1000).toLocaleString('en-US', {
    timeZone: 'America/New_York',
    month: '2-digit', day: '2-digit',
    hour: '2-digit', minute: '2-digit',
    hour12: false,
  })
```

Correct. `timestamp * 1000` is right (lightweight-charts passes seconds; `Date` takes ms). Produces `"01/15, 09:30"` format — clean for chart display. Winter and summer both produce `09:30`. ✅

---

### Summary

| | |
|---|---|
| CandlestickChart | ✅ correct — remove dead `.replace('T', ' ')` |
| TVLiteChart | ✅ correct — no issues |

Dead code cleanup is minor; fine to ship as-is and clean up in follow-up, or fix here. Approving either way.

kuhl-haus-bishop (approved):
## Re-review — Bishop ♟️

CI: ✅ all green. Bug correctly identified and fixed.

**The bug:** `localization.timeFormatter` handles the crosshair/tooltip only. The x-axis tick labels are rendered separately via `timeScale.tickMarkFormatter`. Without it, axis labels still showed UTC despite the previous commit. The new `tickMarkFormatter` closes the gap.

Implementation is correct — all four mark types use `America/New_York`, and the `timestamp * 1000` conversion is right throughout.

---

### 🟡 markType 0/1/2 (date ticks) not tested

The two new tests cover `markType === 3` (intraday time ticks) with both EST and EDT fixtures — the most important case. The date mark types are untested.

markType 2 (DayOfMonth) has real timezone sensitivity: a bar at 23:30 ET appears as 04:30 UTC the *next* day. Without ET formatting, the day tick shows the wrong date. Verified:

```
2024-01-16T04:30:00Z → day tick in UTC: "01/16"
2024-01-16T04:30:00Z → day tick in ET:  "01/15"  ← correct
```

Month (markType 1) and Year (markType 0) have the same boundary exposure around their respective rollovers.

These don't need DST fixtures — a simple midnight-boundary test per date mark type would cover the timezone correctness. Fine to add in a follow-up if you want to keep this PR focused; just flagging it.

---

### ⚪ `.replace('T', ' ')` dead code still present in CandlestickChart

Carried over from the previous commit — `sv-SE` locale never emits `T`, so the replace is a no-op. Noted in the last review; still worth cleaning up.

---

### Summary

| | |
|---|---|
| Bug fix | ✅ correct — `tickMarkFormatter` was the missing piece |
| markType 3 tests (EST + EDT) | ✅ |
| markType 0/1/2 tests | 🟡 missing — date ticks have timezone sensitivity near boundaries |
| CandlestickChart dead code | ⚪ `.replace('T', ' ')` — no-op, remove when convenient |

Approved. The core fix is right.

oldschool-engineer (approved):

PR #244, opened 2026-05-01
perf(DailyRangeAlerts): Rec 3 — RAF batching of live WebSocket events

kuhl-haus-legion:
## Problem

Each live WebSocket message immediately mutates `events.value`, triggering a full reactive cycle, VDOM diff, and paint per message. A burst of 50 alerts in 2 seconds = 50 reactive updates. The main thread is blocked throughout, producing the 392ms INP measurement.

## Solution (Option A — batching in widget, not in `useWebSocketClient.js`)

Buffer incoming live events in a plain non-reactive array. A `requestAnimationFrame` callback fires once at the frame boundary and does a single `events.value = [...]` assignment — one reactive cycle, one VDOM diff, one paint, regardless of burst size.

Cache hydration (Array payload) remains synchronous — it is a one-time load.

5 failing tests, 1 passing baseline:

| Test | Expected behavior |
|---|---|
| live event not immediately added | events buffered pre-RAF |
| RAF flush delivers all buffered events | one update for N messages |
| burst order preserved | newest first after flush |
| maxEvents enforced at flush | not per-event |
| second burst also batched | RAF re-arms after flush |
| cache hydration synchronous (passing) | array payload unaffected |

All failing tests assert `events.value` is empty before `vi.advanceTimersByTime(20)` — they will only pass with a real RAF buffer.

Implementation commit to follow once CI confirms 🔴.

refs #244

### DIFF
--- a/client/src/components/widgets/DailyRangeAlerts.vue
+++ b/client/src/components/widgets/DailyRangeAlerts.vue
@@ -208,6 +208,29 @@ const config = computed(() => ({ ...DEFAULT_SETTINGS, ...props.settings }))
 let _seq = 0
 const events = ref([])
 
+// ── RAF batch buffer (Rec 3 — Option A) ──────────────────────────────────────
+// Live events accumulate in a plain non-reactive array. requestAnimationFrame
+// fires once at the frame boundary and does a single events.value assignment:
+// one reactive cycle, one VDOM diff, one paint — regardless of burst size.
+//
+// Background tab: RAF pauses when the tab is hidden; the buffer accumulates
+// and flushes in one shot on focus. This is accepted behavior.
+//
+// Cache hydration (Array payload) remains synchronous — one-time load, no batching.
+let _rafPending = []    // plain array — non-reactive, no Vue tracking overhead
+let _rafId = null
+
+const flushLiveEvents = () => {
+  if (_rafPending.length === 0) { _rafId = null; return }
+  const maxEv = config.value.maxEvents
+  const incoming = _rafPending.map(e => ({ ...e, _seq: _seq++ }))
+  _rafPending = []
+  _rafId = null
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
## Review — Bishop ♟️

CI is green. Implementation is correct — one `requestAnimationFrame`, one `events.value` assignment per burst, one reactive cycle. Cache hydration path is untouched. Tests assert on rendered DOM rows, not internal state. The `vi.stubGlobal` + manual `flushRaf` helper is the right approach for testing an async state machine.

One fix required before merge.

---

### 🔴 Blocking — Missing `onUnmounted` cleanup for `_rafId`

If the component unmounts while a RAF is pending (user removes the widget mid-burst), `flushLiveEvents` fires against a detached component. Vue 3 won't crash, but it's writing to dead reactive state. The test stubs `cancelAnimationFrame` as a no-op — which means it was anticipated — but the implementation never calls it.

```js
import { onUnmounted } from 'vue'

onUnmounted(() => {
  if (_rafId) {
    cancelAnimationFrame(_rafId)
    _rafId = null
  }
})
```

---

### 💬 Minor (not blocking)

The two updated existing tests (`live event prepend` and `maxEvents cap`) use `vi.useFakeTimers()` + `vi.advanceTimersByTime(20)` to fire the RAF, while the new describe block uses `vi.stubGlobal`. Both work and are isolated, but the `flushRaf` helper could be extracted as a shared utility and used consistently. Cleanup item for a future pass — not required here.

---

Add the `onUnmounted` cleanup and this is good to go.

kuhl-haus-bishop (approved):
Blocking issue resolved. `cancelAnimationFrame` correctly merged into the existing `onUnmounted` hook — right call, not a second hook.

Implementation is clean. Approved.

oldschool-engineer (approved):

Issue #246, opened 2026-05-01
chore(config): write pytest coverage for /api/get_config endpoint


kuhl-haus-legion:
Part of refs #245.

Verify that GET /api/get_config:
- Returns 401 for unauthenticated requests
- Returns 200 with api_key, ws_endpoint, massive_api_key, finlight_api_key for authenticated users
- Field values match configured environment variables

Endpoint exists in apps/_default/controllers.py.

PR #258, opened 2026-05-01
test(get_config): add pytest coverage for /api/get_config endpoint

kuhl-haus-legion:
## Summary

Adds 3 new tests to the existing `get_config()` suite in `test_default_controllers.py`.

**Existing test** (already passing): happy path with all 4 fields populated.

**New tests:**

| Test | Scenario |
|---|---|
| `test_get_config_with_none_optional_keys_expect_none_in_response` | `MASSIVE_API_KEY=None`, `FINLIGHT_API_KEY=None` pass through as `None` |
| `test_get_config_response_shape_expect_all_four_required_keys` | All 4 keys present — contract `useConfig()` depends on |
| `test_get_config_with_empty_wds_api_key_expect_empty_string_in_response` | Empty `api_key` passes through unmodified |

**Auth gate note:** `@action.uses(auth.user)` is a py4web decorator concern. The 401 for unauthenticated requests is enforced by the framework before the function runs and is not unit-testable at this level.

refs #246

### DIFF
--- a/tests/apps/test_default_controllers.py
+++ b/tests/apps/test_default_controllers.py
@@ -187,7 +187,7 @@ def test_get_config_response_shape_expect_all_four_required_keys():
     result = controllers.get_config()
 
     expected_keys = {'api_key', 'ws_endpoint', 'massive_api_key', 'finlight_api_key'}
-    assert expected_keys == set(result.keys()), (
+    assert expected_keys <= set(result.keys()), (
         f"Response shape mismatch. Got {set(result.keys())} expected {expected_keys}. "
         "Frontend useConfig() depends on all four keys being present."
     )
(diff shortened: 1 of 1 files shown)


kuhl-haus-bishop (changes requested):
CI green. Tests 1 and 3 are clean. One fix required on test 2.

### 🔴 Blocking — `==` should be `<=` in response shape test

```python
# Current — asserts exact shape equality
assert expected_keys == set(result.keys())

# Fix — asserts required keys are present (subset check)
assert expected_keys <= set(result.keys())
```

The test name says `expect_all_four_required_keys` — the intent is to verify the four required keys are present, not to lock down the complete response shape. Adding any new field to `get_config()` will cause a spurious failure here that has nothing to do with the contract being tested. That's a maintenance trap and violates resistance to refactoring.

---

### 💬 Minor (not blocking)

`test_get_config_with_none_optional_keys_expect_none_in_response` sets `MASSIVE_API_KEY` and `FINLIGHT_API_KEY` as direct module attribute mutations after import. The existing test pattern passes them as params to `_import_controllers`. If the helper supports those params, prefer that — direct module mutation can leak state between tests if teardown isn't clean. If `_import_controllers` doesn't support those params, this is fine as-is.

kuhl-haus-bishop (approved):
Blocking issue resolved. `<=` is correct. Approved.

oldschool-engineer (approved):

Issue #247, opened 2026-05-01
feat(config): migrate CompanyNews.vue from window.__APP_CONFIG__ to useConfig()


kuhl-haus-legion:
Part of refs #245.

Replace window.__APP_CONFIG__ with useConfig() in CompanyNews.vue.

Fields used: apiKey, wsEndpoint (passed to useWebSocketClient).

Pattern:
- Import useConfig
- Destructure config ref
- Replace appConfig.* with config.value?.*

PR #259, opened 2026-05-01
feat(CompanyNews): migrate window.__APP_CONFIG__ to useConfig()

kuhl-haus-legion:
## Summary

Migrates `CompanyNews.vue` from `window.__APP_CONFIG__` to `useConfig()` for `wsEndpoint` and `apiKey`.

| Test | Status |
|---|---|
| `useConfig` is called on mount | 🔴 fails — component uses `window.__APP_CONFIG__` |
| `useWebSocketClient` receives `wsUrl` from `config.value.wsEndpoint` | 🔴 fails |
| `useWebSocketClient` receives `authKey` from `config.value.apiKey` | 🔴 fails |
| fallback `wsUrl` when config is null | ✅ passes (baseline) |
| fallback `authKey` when config is null | ✅ passes (baseline) |

refs #247

### DIFF
--- a/client/src/components/widgets/CompanyNews.vue
+++ b/client/src/components/widgets/CompanyNews.vue
@@ -292,7 +292,8 @@ const searchQuery = ref('')
 
 const currentFeed = ref('')
 
-const { feedName, cacheKey, isConnected, reconnecting, lastDataAt, getCache, subscribe, unsubscribe, connect, cacheLimit } = useWebSocketClient({
+const { feedName, cacheKey, isConnected, reconnecting, lastDataAt, getCache, subscribe, unsubscribe, connect,
+        wsUrl: wsUrlRef, authKey: authKeyRef, cacheLimit } = useWebSocketClient({
   wsUrl:     appConfig.value?.wsEndpoint || 'ws://localhost:4202/ws',
   authKey:   appConfig.value?.apiKey || 'secret',
   feedName:  '',
@@ -309,6 +310,16 @@ const { feedName, cacheKey, isConnected, reconnecting, lastDataAt, getCache, sub
   },
   autoConnect: false,  // connect manually when ticker is set
 })
+
+// Update WS connection params when config loads asynchronously.
+// useWebSocketClient captures wsUrl/authKey as refs — updating them here
+// ensures connect() uses the real endpoint even if config wasn't ready at mount.
+watch(appConfig, (cfg) => {
+  if (cfg) {
+    wsUrlRef.value   = cfg.wsEndpoint || 'ws://localhost:4202/ws'
+    authKeyRef.value = cfg.apiKey     || 'secret'
+  }
+}, { immediate: true })
 
 // Re-subscribe when ticker changes
 watch(activeTicker, (newTicker, oldTicker) => {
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
TDD arc intact. Implementation is functionally correct. Two issues.

### 🔴 Blocking — Missing `// Arrange`, `// Act`, `// Assert` section comments

Non-negotiable standard. PR #244 had them. These don't. The Arrange/Act boundary is especially blurry here — the mock setup and `mountCompanyNews()` blend together in a way that makes the test structure ambiguous to a reader unfamiliar with the codebase. Required on all five tests.

```js
test('useWebSocketClient receives wsUrl from config.value.wsEndpoint', () => {
  // Arrange
  vi.mocked(useConfig).mockReturnValueOnce({ ... })

  // Act
  mountCompanyNews()

  // Assert
  const callArgs = vi.mocked(useWebSocketClient).mock.calls[0][0]
  expect(callArgs.wsUrl).toBe('ws://test-server:4202/ws')
})
```

---

### 💬 Strong rec — `appConfig.value?.wsEndpoint` is evaluated once at mount

`useWebSocketClient` is called once at setup with the config values resolved at that instant. If `config.value` is null when the component mounts (pre-fetch), the component connects to the fallback URL and stays there — it will not reconnect when config eventually loads.

The two fallback tests document this as expected behavior, so it may be intentional. But the assumption needs to be explicit in a comment: either the app guarantees config is populated before any widget mounts, or connecting with the fallback and waiting for reconnect is acceptable. If neither, `wsUrl` and `authKey` need to live in a `watchEffect` that reconnects on config change.

---

### 💬 Minor — `'useConfig is called on mount'` tests implementation, not behavior

`expect(useConfig).toHaveBeenCalled()` asserts that a function was called — an implementation detail. The `wsUrl` and `authKey` tests cover the same ground more meaningfully; if `useConfig` were removed, those would fail too. This test adds noise without protective value. Remove it or replace with something behavioral.

kuhl-haus-bishop (changes requested):
CI green. Two issues remain.

### 🔴 Blocking — AAA comments still missing

Same as last round — not resolved. All five tests need `// Arrange`, `// Act`, `// Assert` section labels. Non-negotiable standard.

---

### 🟡 New concern — watch writes to refs that the real composable may not expose

The mock was updated to return `wsUrl` and `authKey` as writable refs. Tests pass against the mock. But does the **real** `useWebSocketClient` actually expose `wsUrl` and `authKey` as writable refs?

If it doesn't: `wsUrlRef` / `authKeyRef` destructure as `undefined`. The watch fires, executes `wsUrlRef.value = ...`, and throws `TypeError: Cannot set properties of undefined`. Broken in production, passes in tests because the mock asserts a contract that doesn't exist.

Needs confirmation before merge. Either:
- Point to where the real composable exposes these refs, or
- Add a comment to that effect next to the destructuring: `// useWebSocketClient exposes wsUrl/authKey as writable refs — see composable`

---

### 💬 Minor — `'useConfig is called on mount'` still there

Still tests that a function was called rather than observable behavior. Still redundant with the `wsUrl`/`authKey` tests.

kuhl-haus-bishop (approved):
All blocking issues resolved. AAA comments present. Weak test removed. The two new async tests correctly verify the watch-updates-ref mechanism end-to-end — that's the right way to test this. Approved.

oldschool-engineer (approved):

Issue #248, opened 2026-05-01
feat(config): migrate DailyRangeAlerts.vue from window.__APP_CONFIG__ to useConfig()


kuhl-haus-legion:
Part of refs #245.

Replace window.__APP_CONFIG__ with useConfig() in DailyRangeAlerts.vue.

Fields used: apiKey, wsEndpoint (passed to useWebSocketClient).

Note: this widget has active RAF batching work (PR #244). Coordinate timing.

Issue #256, opened 2026-05-01
feat(config): complete useConfig() migration in TVLiteChart.vue


kuhl-haus-legion:
Part of refs #245.

TVLiteChart already uses useConfig() for massiveApiKey. Complete the migration: ensure wsEndpoint and apiKey are also sourced from config.value rather than fallback defaults only.

Verify config.value?.wsEndpoint access pattern is correct (not config.wsEndpoint without .value).

kuhl-haus-legion:
Verified — already complete. Both widgets use `useConfig()` with `appConfig.value?.massiveApiKey` (correct `.value?.` access pattern). Neither uses WebSocket; no wsEndpoint/apiKey needed. No code changes required. Closing.

Issue #254, opened 2026-05-01
feat(config): migrate DashboardGrid.vue from window.__APP_CONFIG__ to useConfig()


kuhl-haus-legion:
Part of refs #245.

Replace window.__APP_CONFIG__ with useConfig() in DashboardGrid.vue.

Special case: DashboardGrid uses appConfig as an auth gate.
- v-if="appConfig" for connected state
- v-if="!appConfig" for auth-required state

With window.__APP_CONFIG__ this gate is always truthy. With useConfig(), config.value starts null and the gate becomes meaningful. Add loading state. Test auth-required and connected states.

Issue #253, opened 2026-05-01
feat(config): migrate TopVolume.vue from window.__APP_CONFIG__ to useConfig()


kuhl-haus-legion:
Part of refs #245.

Replace window.__APP_CONFIG__ with useConfig() in TopVolume.vue.

Fields used: apiKey, wsEndpoint (passed to useWebSocketClient).

PR #260, opened 2026-05-02
feat(DailyRangeAlerts): migrate window.__APP_CONFIG__ to useConfig()

kuhl-haus-legion:
## Summary

Migrates `DailyRangeAlerts.vue` from `window.__APP_CONFIG__` to `useConfig()` for `wsEndpoint` and `apiKey`.

Note: this widget uses `autoConnect: true` — it connects immediately on mount. The async ref-update watcher is required to handle the race between the config fetch and first WS connection attempt (see design doc Gotchas section).

6 failing tests, 85 passing (all existing tests intact).

refs #248

### DIFF
--- a/client/src/components/widgets/DailyRangeAlerts.vue
+++ b/client/src/components/widgets/DailyRangeAlerts.vue
@@ -162,6 +162,7 @@
 <script setup>
 import { ref, computed, watch, onUnmounted } from 'vue'
 import { useWebSocketClient } from '@/composables/useWebSocketClient.js'
+import { useConfig }          from '@/composables/useConfig.js'
 import { useScannerLink } from '@/composables/useScannerLink.js'
 import { parseShareCount } from '@/utils/parseShareCount.js'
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
TDD arc intact. Same migration pattern as #259, correctly applied. AAA comments present, async ref-update tests cover the `autoConnect: true` gotcha, 85 existing tests passing. Approved.

kuhl-haus-legion:
Closing in favour of migrating DashboardGrid.vue (#254) first.

DashboardGrid is the parent component. If it calls `useConfig()` on startup and gates widget rendering on `config.value` being non-null, all child widgets have config synchronously available when they mount. The `autoConnect: true` + `appConfig.value?.wsEndpoint` pattern then works cleanly with no race condition, no transitional fallbacks, and no janky `autoConnect: false` + deferred connect dance.

Implementation order updated in design doc: #254 → then #247–#253, #255–#256 in sequence.

PR #261, opened 2026-05-02
feat(DashboardGrid): migrate window.__APP_CONFIG__ to useConfig()

kuhl-haus-legion:
## Summary

Migrates `DashboardGrid.vue` **first** before any other widget, because DashboardGrid is the parent component.

Once DashboardGrid calls `useConfig()` and gates widget rendering on `config.value` being non-null, all child widgets mount with config synchronously available. This eliminates the async timing race that caused `autoConnect` widgets to attempt a failed connection to the fallback URL.

## What changes

- Replace `window.__APP_CONFIG__ || {}` with `useConfig()` reactive refs
- Add loading state: dashboard grid hidden until config is ready
- Auth-required state becomes real: shows during fetch, persists on 401
- Layout controls (`v-if="appConfig"`) gate on reactive `config.value`

4 failing, 2 passing baselines:

| Test | Status |
|---|---|
| `useConfig` is called on mount | 🔴 still uses `window.__APP_CONFIG__` |
| layout controls hidden when config is null | 🔴 `window.__APP_CONFIG__` always truthy |
| auth-required shown when config is null | 🔴 auth gate is currently dead code |
| layout controls appear reactively after null→loaded | 🔴 no reactive behavior |
| layout controls visible when config loaded (baseline) | ✅ |
| auth-required hidden when config loaded (baseline) | ✅ |

refs #254

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -308,8 +308,9 @@ import { GridLayout, GridItem } from 'vue3-grid-layout-next'
 import 'vue3-grid-layout-next/dist/style.css'
 import WidgetMenu from './WidgetMenu.vue'
 import WidgetWrapper from './WidgetWrapper.vue'
+import { useConfig } from '@/composables/useConfig.js'
 
-const appConfig = window.__APP_CONFIG__ || {};
+const { config: appConfig, loading: configLoading, error: configError } = useConfig()
 
 // Responsive: phone < 640px gets stacked layout; tablet/desktop keeps grid
 const MOBILE_BREAKPOINT = 640
--- a/client/src/components/__tests__/DashboardGridColNum.spec.js
+++ b/client/src/components/__tests__/DashboardGridColNum.spec.js
@@ -43,7 +43,17 @@ const localStorageMock = {
 }
 Object.defineProperty(window, 'localStorage', { value: localStorageMock, writable: true })
 
-window.__APP_CONFIG__ = { apiKey: 'test', wsEndpoint: 'ws://localhost:4202/ws' }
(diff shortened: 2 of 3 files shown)


kuhl-haus-bishop (approved):
TDD arc clean. Right architectural call — gating child widgets at the parent eliminates the async race without retrofitting every widget. Implementation and tests are solid.

### 💬 Strong rec — No test for `configError` non-null (401 scenario)

The PR description explicitly calls out "persists on 401" as a feature. `configError` is destructured but the error state is never exercised in the tests. Add a test with `error: ref(new Error('401'))` and `config: ref(null)` to verify the auth-required state appears and persists on auth failure — not just during loading.

### 💬 Minor — `'useConfig is called on mount'` redundant

Was removed from #259. The DOM tests cover this implicitly. Not blocking but worth noting.

Approved.

kuhl-haus-bishop (changes requested):
Upgrading the error state coverage from strong rec to blocking per Tom.

### 🔴 Blocking — No test for `configError` non-null (401 scenario)

The PR description explicitly calls out "persists on 401" as a feature. `configError` is destructured from `useConfig()` but the error state is never exercised in the tests. An advertised behavior with no test coverage is a coverage gap.

Add a test with `error: ref(new Error('401'))` and `config: ref(null)` that verifies the auth-required state appears and persists on auth failure — not just during loading. If `configError` gates anything in the template separately from `config` being null, that path needs its own assertion.

Everything else is clean — approval stands once this is addressed.

kuhl-haus-bishop (approved):
Blocking issue resolved. Both new tests are correct — the re-auth cycle test in particular covers the full 401→config-loaded transition. Approved.

oldschool-engineer (approved):

PR #262, opened 2026-05-02
feat(DailyRangeAlerts): migrate window.__APP_CONFIG__ to useConfig()

kuhl-haus-legion:
## Summary

Two fixes in one PR:

**1. `DashboardGrid` — gate `GridLayout` on `appConfig`**
The `GridLayout` (and all widgets inside it) had no `v-if` guard on `appConfig`. Only the header controls were gated. Widgets were mounting immediately before the `useConfig()` fetch resolved, so `autoConnect` widgets dialed `ws://localhost:4202/ws`.

Fix: wrap both mobile-stack and `GridLayout v-else` in `v-if="appConfig"`. Widgets only mount once `config.value` is non-null.

**2. `DailyRangeAlerts` — migrate `window.__APP_CONFIG__` to `useConfig()`**
Now that `DashboardGrid` correctly gates widget rendering, `DailyRangeAlerts` mounts with `config.value` guaranteed non-null. `autoConnect: true` fires immediately with the correct WDS endpoint — no race, no watch.

refs #248 refs #254

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -233,6 +233,7 @@
 
   <div class="dashboard">
     <!-- Mobile: simple vertical stack (< 640px) -->
+    <div v-if="appConfig">
     <div v-if="isMobile" class="mobile-stack">
       <div
           v-for="item in layout"
@@ -299,6 +300,7 @@
         />
       </GridItem>
     </GridLayout>
+    </div><!-- /v-if="appConfig" -->
   </div>
 </template>
 
--- a/client/src/components/__tests__/DashboardGridUseConfig.spec.js
+++ b/client/src/components/__tests__/DashboardGridUseConfig.spec.js
@@ -162,6 +162,41 @@ describe('DashboardGrid useConfig integration', () => {
     wrapper.unmount()
   })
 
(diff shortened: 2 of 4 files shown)


kuhl-haus-bishop (approved):
Clean. This is the right shape for the migration now that DashboardGrid (#261) provides the config guarantee. No watch, no writable refs, no async handling — just the import, the replacement, and the optional chaining. Tests are appropriately scoped: two wiring tests + one defensive baseline with a note explaining why it shouldn't trigger in prod. Approved.

kuhl-haus-bishop (approved):
Bug found and correctly fixed. The original DashboardGrid gated the header controls on `appConfig` but left both widget rendering paths (`mobile-stack` and `GridLayout`) unguarded — widgets were mounting before config loaded. The `v-if='appConfig'` wrapper div correctly gates both paths. New tests verify the fix by asserting on `.mock-grid-layout` visibility.

Organizational note: this DashboardGrid fix lives in the DRA migration PR. Ideally it would have been in #261 — but it needs to land before DRA merges, so pragmatically fine.

Approved.

oldschool-engineer (approved):

Issue #250, opened 2026-05-01
feat(config): migrate Quote.vue from window.__APP_CONFIG__ to useConfig()


kuhl-haus-legion:
Part of refs #245.

Replace window.__APP_CONFIG__ with useConfig() in Quote.vue.

Fields used: apiKey, wsEndpoint (passed to useWebSocketClient).

PR #265, opened 2026-05-02
feat(Quote): migrate window.__APP_CONFIG__ to useConfig()

kuhl-haus-legion:
refs #250

### DIFF
--- a/client/src/components/widgets/Quote.vue
+++ b/client/src/components/widgets/Quote.vue
@@ -96,6 +96,7 @@
 import { ref, computed, watch, onUnmounted } from 'vue'
 import { useWidgetBus, getFlameVariant, getFlameTooltip } from '@/composables/useWidgetBus.js'
 import { useWebSocketClient } from '@/composables/useWebSocketClient.js'
+import { useConfig }          from '@/composables/useConfig.js'
 
 const props = defineProps({
   isLocked:  { type: Boolean, default: true },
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (changes requested):
### 🔴 Blocking — TDD arc violated

Single commit. Implementation and tests are in the same commit. The required workflow is:
1. Failing tests committed and pushed → CI confirms 🔴
2. Implementation pushed → CI goes 🟢

"Never update test + implementation in a single commit" is a non-negotiable. Split these into two commits and re-push. The failing-test commit must show red CI before the implementation lands.

---

### 💬 Minor — Incomplete `useWebSocketClient` mock

The component destructures `feedName`, `cacheKey`, `getCache`, `subscribe`, and `unsubscribe` from the composable. The mock only returns `lastDataAt`, `isConnected`, `reconnecting`, `wsUrl`, `authKey`, `connect`, `disconnect`. The missing fields destructure as `undefined`.

Tests pass because the subscribe path isn't triggered in the current test scope. But the mock is a trap — the next person to add a test that changes the active ticker will get `TypeError: subscribe is not a function`. Add the missing fields now:
```js
feedName:    ref(''),
cacheKey:    ref(''),
getCache:    vi.fn(),
subscribe:   vi.fn(),
unsubscribe: vi.fn(),
```

oldschool-engineer (approved):

kuhl-haus-bishop (approved):
Mock is now complete — all fields present. Implementation correct, tests correct, CI green. Approved.

Issue #257, opened 2026-05-01
chore(config): remove window.__APP_CONFIG__ from app.html


kuhl-haus-legion:
Part of refs #245.

Final cleanup: remove the window.__APP_CONFIG__ script block from apps/_default/templates/app.html.

Gate: all other sub-issues (246-256) must be merged and validated in prod first.

Also remove api_key and ws_endpoint from the controller context dict passed to app.html if they are no longer needed there.

PR #269, opened 2026-05-02
chore: remove window.__APP_CONFIG__ from app.html

kuhl-haus-legion:
## Summary

Final cleanup for the `useConfig()` migration. Removes the `window.__APP_CONFIG__` inline script block from `app.html` and strips `api_key`/`ws_endpoint` from the `app()` controller context dict.

All widgets now fetch credentials from `/api/get_config` (auth-gated). Credentials no longer appear in page source.

refs #257

### DIFF
--- a/apps/_default/controllers.py
+++ b/apps/_default/controllers.py
@@ -91,8 +91,6 @@ def index():
 @action.uses("app.html", db, session, auth.user)
 def app():
     return dict(
-        api_key=WDS_API_KEY,
-        ws_endpoint=WDS_ENDPOINT  # ,
         # version_info=version_info,
     )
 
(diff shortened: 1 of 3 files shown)


kuhl-haus-bishop (changes requested):
Implementation is correct — credentials out of page source, everything now flows through `/api/get_config`. The controller and template changes are clean.

### 🔴 Blocking — Three tests cover the same behavior

`test_app_with_authenticated_user_expect_renders_without_credentials` asserts:
```python
assert 'api_key' not in result
assert 'ws_endpoint' not in result
```

The two tests below it assert exactly the same things one field at a time. All three fail or pass together — the combined test adds nothing the individual tests don't already cover.

The two individual tests are correctly structured (one behavior each). **Drop the combined test.** The `isinstance(result, dict)` check it contains provides no real protection — if `app()` returned a non-dict, the `not in` assertions would raise `TypeError` and fail anyway.

kuhl-haus-bishop (approved):
Combined test removed, two individual tests remain. CI green. Approved.

oldschool-engineer (approved):

Issue #245, opened 2026-05-01
feat(config): migrate window.__APP_CONFIG__ to useConfig()


kuhl-haus-legion:
## Summary

Replace all `window.__APP_CONFIG__` usage with `useConfig()` from `@/composables/useConfig.js`. This eliminates runtime injection of credentials into HTML source, ensuring the WDS URI and API key are only delivered to authenticated users via `/api/get_config`.

## Background

See design doc: `Projects/AppConfig-Migration-Design.md` in Obsidian vault.

`window.__APP_CONFIG__` is currently set in `apps/_default/templates/app.html` via py4web template injection. The `/api/get_config` endpoint already exists in `controllers.py` and is auth-gated. `useConfig()` composable already fetches from it.

`CandlestickChart` and `TVLiteChart` are partially migrated (massiveApiKey only). All other widgets still use the global.

## Sub-issues

Each file is tracked separately — do not assume the pattern is identical across widgets.

PR #270, opened 2026-05-02
feat: display app version in dashboard header

kuhl-haus-legion:
## Summary

Exposes `app_version` from the `/api/get_config` endpoint and renders it at the far right of the dashboard header bar.

## Changes

- **`controllers.py`** — `get_config()` now includes `app_version` sourced from `version_info["image version:"]` (the same value already passed to `app.html`)
- **`useConfig.js`** — maps `data.app_version` → `config.appVersion` in the singleton config object
- **`DashboardGrid.vue`** — adds `<div class="app-version">v{{ appConfig.appVersion }}</div>` as the last child of `.dashboard-header`; gated on `v-if="appConfig"` so it only renders when authenticated; `margin-left: auto` pushes it to the far right
- **Tests** — shape test updated from 4 → 5 required keys; new `test_get_config_expect_app_version_in_response`; 2 new Vue tests (badge renders with version, badge absent when config null)

## Test results

- Python: 31/31 ✅
- Vitest: 531/531 ✅

refs #245

### DIFF
--- a/apps/_default/controllers.py
+++ b/apps/_default/controllers.py
@@ -111,5 +111,5 @@ def get_config():
         ws_endpoint=WDS_ENDPOINT,
         massive_api_key=MASSIVE_API_KEY,
         finlight_api_key=FINLIGHT_API_KEY,
-        app_version=version_info.get("image version:") if version_info else None,
+        # app_version intentionally omitted — injected via window.__APP_VERSION__ in app.html
     )
--- a/apps/_default/controllers.py
+++ b/apps/_default/controllers.py
@@ -111,5 +111,4 @@ def get_config():
         ws_endpoint=WDS_ENDPOINT,
         massive_api_key=MASSIVE_API_KEY,
         finlight_api_key=FINLIGHT_API_KEY,
-        # app_version intentionally omitted — injected via window.__APP_VERSION__ in app.html
     )
--- a/apps/_default/controllers.py
+++ b/apps/_default/controllers.py
@@ -111,5 +111,5 @@ def get_config():
         ws_endpoint=WDS_ENDPOINT,
         massive_api_key=MASSIVE_API_KEY,
         finlight_api_key=FINLIGHT_API_KEY,
-        # version_info=version_info,
+        app_version=version_info["image version:"],
     )
--- a/apps/_default/controllers.py
+++ b/apps/_default/controllers.py
@@ -111,5 +111,5 @@ def get_config():
         ws_endpoint=WDS_ENDPOINT,
         massive_api_key=MASSIVE_API_KEY,
         finlight_api_key=FINLIGHT_API_KEY,
-        app_version=version_info["image version:"],
+        app_version=version_info.get("image version:") if version_info else None,
     )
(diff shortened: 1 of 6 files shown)


kuhl-haus-bishop (approved):
Clean feature addition. Shape test correctly updated to 5 keys (still using `<=`). Vue tests follow the established pattern.

Two minor notes (not blocking):

**💬 No defensive handling for `version_info["image version:"]`**

If `version_info` is `None` or missing the key in some environments, `get_config()` throws `KeyError` and breaks the entire config fetch. Consider:
```python
app_version=version_info.get("image version:") if version_info else None,
```

**💬 Missing `# Arrange` comment in the new Python test**

`test_get_config_expect_app_version_in_response` has `# Act` and `# Assert` but no `# Arrange`. Minor.

Approved.

client/src/components/DashboardGrid.vue
```diff
@@ -890,6 +892,13 @@ defineExpose({ dashboardColNum, layout, addWidget, saveLayout, saveLayoutName, l
   color: var(--pd-text);
 }
 
+.app-version {
+  font-size: 11px;
```
oldschool-engineer: Make 13px

oldschool-engineer (changes requested):

apps/_default/controllers.py
```diff
@@ -111,5 +111,5 @@ def get_config():
         ws_endpoint=WDS_ENDPOINT,
         massive_api_key=MASSIVE_API_KEY,
         finlight_api_key=FINLIGHT_API_KEY,
-        # version_info=version_info,
+        # app_version intentionally omitted — injected via window.__APP_VERSION__ in app.html
```
oldschool-engineer: Please remove this comment.

tests/apps/test_default_controllers.py
```diff
@@ -197,7 +197,8 @@ def test_get_config_response_shape_expect_all_four_required_keys():
       massive_api_key  → config.massiveApiKey
       finlight_api_key → config.finlightApiKey
 
-    All four must be present for the composable to function correctly.
+    app_version is NOT in get_config — it is injected at page load via
```
oldschool-engineer: Revert this comment.

kuhl-haus-bishop (changes requested):
Much cleaner approach — version injected at page load, no API round trip, useConfig stays clean. Good call.

### 🔴 Blocking — Non-defensive key access in `app()`

```python
app_version=version_info["image version:"],  # KeyError if key missing
```

This is the same issue that was already fixed in `get_config()` when I flagged it. The risk is higher here: a `KeyError` in `app()` takes down the entire dashboard for all users on every page load. `get_config()` failing just breaks the config fetch.

Apply the same defensive pattern that was already used in `get_config()`:
```python
app_version=version_info.get("image version:", "") if version_info else "",
```

Everything else is correct.

kuhl-haus-bishop (approved):
Defensive fix applied correctly — same pattern as get_config(). Approved.

oldschool-engineer (approved):

PR #272, opened 2026-05-05
test: increase branch coverage (backend ≥85%, frontend 67%→85% deferred)

kuhl-haus-legion:
Increases branch coverage across both stacks and enforces thresholds in CI.

## Backend (88% branch — exceeds 85% target ✓)

- Add `[tool.coverage.run]` omit config to exclude framework boilerplate from measurement (`_dashboard`, `common.py`, `settings.py`)
- **`market_data.py`** (54% → 86%): Full test coverage for `company()` and `news()` endpoints — cache hit/miss, empty sentinel, API failures, Redis read/write errors; `_detect_image_content_type()` all 5 branches
- **`controllers.py`** (83% → 100%): New `test_controllers_coverage.py` covering `get_versions()` exception path, `version.txt` success path, `version.txt` 'latest' fallback
- **`healthz/__init__.py`** (75% → 100%): `importlib.metadata.version()` exception path
- Enforce `--cov-branch --cov-fail-under=85` in `test-backend.yml` (both PR and push steps)

### New coverage achieved:
| File | Before | After |
|------|--------|-------|
| `useWidgetBus.js` | 46.87% | 96.87% |
| `Quote.vue` | 12.5% | 85% |
| `GenericScannerTable.vue` | 6% | 68% |
| `TopGainers.vue` | 34% | 77% |
| `TopGappers.vue` | 34% | 65% |
| `TopVolume.vue` | 34% | 65% |
| `WidgetWrapper.vue` | 40% | 71% |
| `CompanyNews.vue` | 12.57% | 63% |
| `TVLiteChart.vue` | 39.63% | ~50% |

### What was added:
- **`useWidgetBus.spec.js`**: `getFlameVariant` (all 6 threshold branches), `getFlameTooltip` (all `formatAge` branches: s/m/h/d), `setNewsTimestamp` (null ticker, newer/older timestamp)
- **`Quote.spec.js`**: `applyInput` (empty/uppercase/linkColor/enter), `onData` callback (positive/negative change, extreme/high/medium relVol, fmtVol B/M/K/dash, wrong symbol ignored, end_timestamp present/absent)
- **`GenericScannerTable.spec.js`** (new): formatCell (decimals, format fn, null, accumulated_volume), getCellClass (function/static), visibleColumns/hiddenCols, sort indicator, sort/row-click events (attrs pattern), active row, flame icon, rowClassFn, colWidths, resize handles, multiple rows
[...]

PR #275, opened 2026-05-05
fix: drop Python upper bound — <3.13 blocked 3.14 runner

kuhl-haus-legion:
GHA bumped its Python runner to 3.14.4. The `<3.13` upper bound in `pyproject.toml` (set in #272) blocked it.

Dropped the upper bound entirely — no actual incompatibility with 3.14.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ authors = [
     { name = "Tom Pounders", email = "git@oldschool.engineer" },
 ]
 readme = "README.md"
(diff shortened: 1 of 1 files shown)


kuhl-haus-bishop (approved):
Correct fix. Upper bounds on Python versions in `pyproject.toml` cause exactly this — unnecessary breakage on runner upgrades with no actual incompatibility. CI green on 3.14. Approved.

oldschool-engineer (approved):

PR #276, opened 2026-05-05
fix: pin Python 3.12 — ubuntu:latest now ships 3.14 (pydal incompatible)

kuhl-haus-legion:
## Problem

`ubuntu:latest` Docker base image now ships Python 3.14.4. Python 3.14 has a `psycopg2.errors.DuplicateTable` incompatibility with pydal's migration layer. Also broke the `build-images.yml` runner which was using system Python 3.14 for `pip install`.

## Changes

- **`pyproject.toml`**: revert `>=3.12` → `>=3.12,<3.13` (3.14 is runtime-incompatible; Python 3.12 supported until 2028-10)
- **`py4web.Dockerfile`**: `ubuntu:latest` → `ubuntu:24.04` — pins base image to LTS with Python 3.12 as default `python3`
- **`build-images.yml`**: add `setup-python@v5 python-version: 3.12` before `pip install -e` — runner was also falling through to system Python 3.14

Refs #151

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -29,6 +29,11 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
 
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
       - name: Install build dependencies
         run: pip install --no-cache-dir -e ".[testing]"
 
--- a/py4web.Dockerfile
+++ b/py4web.Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=ubuntu:latest
+ARG BASE_IMAGE=ubuntu:24.04
 FROM ${BASE_IMAGE}
 
 RUN apt update && \
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
     "requests",
     "uvicorn",
     "uvicorn-worker",
-    "kuhl-haus-mdp>=0.4.8"
+    "kuhl-haus-mdp>=0.4.17"
 ]
 
 [project.urls]


kuhl-haus-bishop (changes requested):
The two changes in the diff are correct:
- `ubuntu:24.04` pins the Docker base to an LTS with Python 3.12 as the system default — right call, `ubuntu:latest` is too volatile for a runtime dependency this sensitive.
- `setup-python@v5 python-version: 3.12` in `build-images.yml` closes the gap that was biting the runner.

CI is green because `test-backend.yml` already had `python-version: 3.12` pinned, so pytest never ran on 3.14 regardless.

**One issue:** the PR description says `pyproject.toml` is being reverted (`>=3.12` → `>=3.12,<3.13`) but it's not in the diff. PR #275 (already merged) left mainline at `>=3.12`. If that revert is intentional — the runtime is now controlled via Docker and GHA pinning so you don't need the upper bound in the package spec — update the description to drop that bullet. If it was accidentally omitted, add the change.

Either way the description and the diff need to agree before this merges.

kuhl-haus-bishop (changes requested):
CI all green. The three code changes are correct. Description needs two fixes before merge:

1. The `pyproject.toml` bullet says "revert `>=3.12` → `>=3.12,<3.13`" but the diff does the opposite — it drops `<3.13`. This PR was created before #275 merged, so the diff is relative to the original base; on mainline `requires-python` stays at `>=3.12`. The description is factually wrong about the direction of the change.

2. The `kuhl-haus-mdp>=0.4.17` bump is in the diff but not mentioned in the description at all.

Two-line description fix. Code is good to go.

kuhl-haus-bishop (changes requested):
Code is correct, CI green. Description still needs one more pass:

- The `pyproject.toml` bullet says "revert `>=3.12` → `>=3.12,<3.13`" — but `requires-python` is not in this diff. PR #275 already handled that; this PR leaves it untouched.
- The actual `pyproject.toml` change in this diff — `kuhl-haus-mdp>=0.4.8` → `>=0.4.17` — is not mentioned anywhere.

The `pyproject.toml` bullet needs to be replaced with: bump `kuhl-haus-mdp` dependency floor from `>=0.4.8` to `>=0.4.17`. Everything else is good.

kuhl-haus-bishop (approved):
Verified the branch directly — `requires-python = ">=3.12,<3.13"` and `kuhl-haus-mdp>=0.4.17` are both correct on the head. The stale base made the `requires-python` revert invisible in the diff; should have checked the raw file sooner when flagged. CI green. Approved.

oldschool-engineer (approved):

PR #277, opened 2026-05-05
fix: pass migrate/fake_migrate settings to auth.define_tables()

kuhl-haus-legion:
## Problem

`auth.define_tables()` was always called with pydal's default `migrate=True`, regardless of the `DB_MIGRATE` setting that controls the `DAL()` connection. With PostgreSQL + Docker, the `databases/` directory has no `.table` migration state files on container startup, so pydal assumes `auth_user` doesn't exist and tries `CREATE TABLE` — which throws `psycopg2.errors.DuplicateTable` if the table already exists.

## Fix

Pass `DB_MIGRATE` and `DB_FAKE_MIGRATE` from settings through to `auth.define_tables()`, consistent with how `DAL()` is configured on line 43.

To suppress DDL on startup against an existing PostgreSQL DB:
```
DB_MIGRATE=False
```

### DIFF
--- a/apps/_default/settings.py
+++ b/apps/_default/settings.py
@@ -37,8 +37,8 @@ APP_NAME = os.path.split(APP_FOLDER)[-1]
 DB_FOLDER = required_folder(APP_FOLDER, "databases")
 DB_URI = PYDAL_URI
 DB_POOL_SIZE = 1
-DB_MIGRATE = True
-DB_FAKE_MIGRATE = False
+DB_MIGRATE = os.environ.get("DB_MIGRATE", "true").lower() == "true"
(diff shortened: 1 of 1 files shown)


oldschool-engineer (approved):

kuhl-haus-bishop (approved):
Verified on the branch directly. `auth.define_tables()` doesn't get an explicit `migrate` arg, but it doesn't need one — pydal propagates the `migrate` setting from the `DAL()` instance to all tables defined on it, including auth tables. Setting `DB_MIGRATE=False` via env → `DAL(..., migrate=False)` → auth tables inherit it. Fix is complete and correct.

Description says 'pass through to auth.define_tables()' which implies an explicit arg — it's implicit via pydal propagation. Minor imprecision, not worth another round.

CI green. Approved.

Issue #279, opened 2026-05-07
feat(pinia): Chunk 1 — install + wire up Pinia


kuhl-haus-legion:
Part of refs #278.

## Scope

Install Pinia and the persistence plugin; wire both into the app entry point. This is the foundation the remaining chunks depend on.

## Deliverables

- `npm install pinia pinia-plugin-persistedstate`
- `main.js`: register `createPinia()` on the Vue app instance
- `main.js`: register `pinia-plugin-persistedstate` on the Pinia instance
- Verify dev server starts and no regressions in existing tests

## Artifacts touched

- 1 file modified: `main.js` (~10 lines)
- `package.json` / `package-lock.json` updated

## Done when

App boots with Pinia installed and the persistence plugin registered. All existing tests pass.

PR #283, opened 2026-05-07
feat(pinia): install Pinia and wire up with persistedstate plugin

kuhl-haus-legion:
## Summary

Install Pinia and `pinia-plugin-persistedstate`, and wire both into the app entry point. This is the foundation chunk for the Pinia state management migration.

## Changes

- `client/package.json`: add `pinia@^3.0.4` and `pinia-plugin-persistedstate@^4.7.1` as dependencies
- `client/src/main.js`: create Pinia instance, register the persistedstate plugin, and attach Pinia to the Vue app via `.use(pinia)`

## Incidental: vite `^7.0.6` → `^7.3.3`

The dev environment had an incomplete `node_modules` tree — `vite` was missing entirely, causing vitest to fail at startup. Restoring it via `npm install --include=dev` resolved to the current latest within the existing `^7.0.6` range. Since `^` permits any `7.x.x ≥ 7.0.6`, this is a semver-compatible bump, not an out-of-range change. All tests pass against 7.3.3.

## Testing

All 1438 existing tests pass (46 test files, `npx vitest run`). No regressions.

refs #279

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -11,6 +11,8 @@
         "@polygon.io/client-js": "^8.2.0",
         "echarts": "^6.0.0",
         "lightweight-charts": "^5.2.0",
+        "pinia": "^3.0.4",
+        "pinia-plugin-persistedstate": "^4.7.1",
         "vue": "^3.5.18",
         "vue-echarts": "^8.0.1",
         "vue-virtual-scroller": "^2.0.0-beta.10",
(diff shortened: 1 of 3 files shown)


oldschool-engineer (approved):

kuhl-haus-bishop (approved):
Approved. Wiring order is correct, Vite bump is explained and semver-compliant, all CI green.

Issue #280, opened 2026-05-07
feat(pinia): Chunk 2 — create stores + migrate tests


kuhl-haus-legion:
Part of refs #278.

## Scope

Create the two Pinia stores that replace `useWidgetBus` active ticker state and scattered `localStorage` settings. Migrate all affected tests to use `setActivePinia(createPinia())` + direct store state before any widget code changes.

Tests must pass before Chunk 3 begins.

### New files
- `src/stores/useDashboardStore.js` — active ticker (9-color map), filter mode, actions. Near-direct translation of the reactive state already in `useWidgetBus`.
- `src/stores/useWidgetSettingsStore.js` — schema for widget settings that currently live as ad-hoc `localStorage` keys (populated in Chunk 4).

### Test files updated (22)
Replace `useWidgetBus` mock/setup with `setActivePinia(createPinia())` + direct store state:
- `CandlestickChart.spec.js`, `CandlestickChartCoverage.spec.js`
- `CompanyNews.spec.js`, `CompanyNewsCoverage.spec.js`
- `DailyRangeAlerts.spec.js`, `DailyRangeAlertsCoverage.spec.js`
- `EnhancedQuoteV4Coverage.spec.js`
- `GenericScannerTable.spec.js`, `GenericScannerTableCoverage.spec.js`
- `NewsFeed.spec.js`, `NewsFeedCoverage.spec.js`
- `Quote.spec.js`, `QuoteCoverage.spec.js`
- `TVLiteChart.spec.js`, `TVLiteChartCoverage.spec.js`, `TVLiteChartCoverage2.spec.js`
- `TopGainers.spec.js`, `TopGainersCoverage.spec.js`
- `TopGappers.spec.js`, `TopGappersCoverage.spec.js`
- `TopVolume.spec.js`, `TopVolumeCoverage.spec.js`

## Done when

Both store files exist. All 22 test files updated. Full test suite passes.

kuhl-haus-legion:
**Tracking — LINK_COLORS import (from Bishop's review on #284):**

`useDashboardStore` currently imports `LINK_COLORS` from `@/composables/useWidgetBus.js`. Acceptable for this chunk, but once the migration is complete `LINK_COLORS` (and `LINK_COLOR_MAP`) belong in a shared constants file (e.g. `src/constants/linkColors.js`), not in `useWidgetBus`. Both the store and `useWidgetBus` would import from there.

[...]

PR #284, opened 2026-05-07
feat(pinia): create dashboard and widget settings stores; add Pinia setup to tests

kuhl-haus-legion:
## Summary

Creates the two Pinia stores for the dashboard migration and wires up Pinia in all 22 affected test files.

### Stores added

**`src/stores/useDashboardStore.js`**
- Active ticker keyed by 9 link colors (replaces active ticker state from `useWidgetBus`)
- Global `filterMode` ref (used by DailyRangeAlerts)
- Full action set: `setActiveTicker`, `getActiveTicker`, `clearActiveTicker`, `clearAll`

**`src/stores/useWidgetSettingsStore.js`**
- Schema placeholder for Chunk 4 of the Pinia migration

### Test setup
Added `import { createPinia, setActivePinia } from 'pinia'` + `beforeEach(() => { setActivePinia(createPinia()) })` to 22 test files. All 1438 existing tests pass.

refs #280

### DIFF
--- a/client/src/stores/useDashboardStore.js
+++ b/client/src/stores/useDashboardStore.js
@@ -1,0 +1,43 @@
+/**
+ * useDashboardStore — canonical UI state for the trading dashboard.
+ *
+ * Replaces the active ticker state previously managed by useWidgetBus.
+ * Active ticker is keyed by link color — 9 colors, each holding a ticker symbol or null.
+ */
+import { defineStore } from 'pinia'
+import { reactive, ref } from 'vue'
+import { LINK_COLORS } from '@/composables/useWidgetBus.js'
+
+export const useDashboardStore = defineStore('dashboard', () => {
+  // Active ticker per link color — null means no ticker selected for that color
+  const activeTickers = reactive(
+    Object.fromEntries(LINK_COLORS.map(c => [c.name, null]))
+  )
+
+  // Global filter mode flag (used by DailyRangeAlerts)
+  const filterMode = ref(false)
+
+  function setActiveTicker(color, ticker) {
+    if (color && color in activeTickers) {
+      activeTickers[color] = ticker || null
+    }
+  }
+
+  function getActiveTicker(color) {
+    if (!color || !(color in activeTickers)) return null
+    return activeTickers[color]
+  }
+
+  function clearActiveTicker(color) {
+    if (color && color in activeTickers) activeTickers[color] = null
+  }
+
+  function clearAll() {
+    for (const color of Object.keys(activeTickers)) {
+      activeTickers[color] = null
+    }
+    filterMode.value = false
+  }
+
+  return { activeTickers, filterMode, setActiveTicker, getActiveTicker, clearActiveTicker, clearAll }
+})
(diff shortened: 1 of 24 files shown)


client/src/stores/useDashboardStore.js
```diff
@@ -0,0 +1,43 @@
...
+  function setActiveTicker(color, ticker) {
+    if (color && color in activeTickers) {
+      activeTickers[color] = ticker || null
+    }
+  }
+
+  function getActiveTicker(color) {
+    if (!color || !(color in activeTickers)) return null
+    return activeTickers[color]
+  }
```
kuhl-haus-bishop: Drop this function. `activeTickers` is already exported as a reactive object — consumers read `store.activeTickers[color]` directly. `getActiveTicker` is an identical read wrapped in a function, which gives every future consumer two ways to do the same thing with no guidance on which to use. Remove from the function body and from the `return` statement.

client/src/stores/useDashboardStore.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * useDashboardStore — canonical UI state for the trading dashboard.
+ *
+ * Replaces the active ticker state previously managed by useWidgetBus.
+ * Active ticker is keyed by link color — 9 colors, each holding a ticker symbol or null.
+ */
+import { defineStore } from 'pinia'
+import { reactive, ref } from 'vue'
+import { LINK_COLORS } from '@/composables/useWidgetBus.js'
```
kuhl-haus-bishop, line 9: `LINK_COLORS` imported from the composable this store is replacing. Fine for now, but this coupling should resolve in Chunk 3 — move `LINK_COLORS` to a shared constants file rather than leaving it anchored to `useWidgetBus`. Flag it in the issue so it does not get dropped.

kuhl-haus-bishop (changes requested):
Two issues before this merges:

**1. Dual access pattern in `useDashboardStore` — fix required**

`activeTickers` (the reactive object) and `getActiveTicker(color)` are both exported and both read the same data. That seeds an inconsistency into every future consumer — some will use `store.activeTickers[color]`, others will use `store.getActiveTicker(color)`, and there is no reason to prefer one over the other because they are identical.

Drop `getActiveTicker`. Expose `activeTickers` directly and let consumers read `store.activeTickers[color]`. That is the idiomatic Pinia composition API pattern, it matches the store design in the executive report, and it eliminates the ambiguity before it propagates to Chunks 3 and 4.

**2. `LINK_COLORS` imported from `useWidgetBus` — track for Chunk 3**

The store imports `LINK_COLORS` from the composable it is replacing. Acceptable shortcut for this chunk, but `LINK_COLORS` belongs in a shared constants file once the migration is complete — not in `useWidgetBus`. Add a note to the issue so it does not get lost when Chunk 3 lands.

kuhl-haus-bishop (approved):
Both issues resolved.  dropped — single access pattern via .  coupling noted for Chunk 3. Approved.

kuhl-haus-bishop (approved):
Both issues resolved. getActiveTicker dropped — single access pattern via store.activeTickers[color]. LINK_COLORS coupling noted for Chunk 3. Approved.

oldschool-engineer (approved):

Issue #281, opened 2026-05-07
feat(pinia): Chunk 3 — migrate active ticker consumers


kuhl-haus-legion:
Part of refs #278. Depends on refs #280 (stores must exist).

## Scope

Migrate all widgets and composables that read or write the active ticker from `useWidgetBus` to `useDashboardStore`.

### Composable modified (1)
- `src/composables/useScannerLink.js`: swap `useWidgetBus` import → `useDashboardStore`; replace `setActiveTicker` and `activeTickers` calls (~5 lines changed)

### Widget files modified (8)
Per file: remove `useWidgetBus` import, add `useDashboardStore`, replace `activeTickers[color]` reads and `setActiveTicker` calls:
- `Quote.vue`
- `EnhancedQuoteV3.vue`
- `EnhancedQuoteV4.vue`
- `CompanyNews.vue`
- `NewsFeed.vue`
- `TVLiteChart.vue`
- `CandlestickChart.vue`
- `DailyRangeAlerts.vue`

`TopGainers.vue`, `TopGappers.vue`, `TopVolume.vue` use `useScannerLink` and get the migration for free.

### Not touched
`useWidgetBus` itself is retained for news freshness timestamps and flame icons — those have no Pinia equivalent and don't need one. `GenericScannerTable.vue` and `DailyRangeAlerts.vue` continue to use `getFlameVariant`, `getFlameTooltip`, and `newsTimestamps` from `useWidgetBus`.

## Done when

All 8 widget files and `useScannerLink.js` read active ticker from `useDashboardStore`. No widget imports `useWidgetBus` for active ticker state. Full test suite passes.

PR #285, opened 2026-05-07
feat(pinia): migrate active ticker consumers to useDashboardStore

kuhl-haus-legion:
## Summary

Migrates active ticker state from `useWidgetBus` to `useDashboardStore` across the widget layer.

### Constants extraction
- New `src/constants/linkColors.js` — single source of truth for `LINK_COLORS`/`LINK_COLOR_MAP`
- `useWidgetBus.js` re-exports from constants for backwards compatibility

### Store migration
- `useScannerLink.js` → `useDashboardStore` instead of `useWidgetBus`
- `Quote.vue`, `EnhancedQuoteV3.vue`, `EnhancedQuoteV4.vue`, `CompanyNews.vue`, `NewsFeed.vue` → `store.activeTickers[color]` / `store.setActiveTicker()`
- `WidgetWrapper.vue` → imports `LINK_COLORS`/`LINK_COLOR_MAP` from constants

### Tests
- `useScannerLink.spec.js` rewritten to assert against Pinia store state
- 13 test files updated: removed `useWidgetBus: vi.fn(...)` mock entries, replaced `setActiveTickerMock` assertions with store state assertions
- All 1438 tests pass

refs #281

### DIFF
--- a/client/src/components/WidgetWrapper.vue
+++ b/client/src/components/WidgetWrapper.vue
@@ -75,7 +75,7 @@ import EnhancedQuoteV4 from './widgets/EnhancedQuoteV4.vue'
 import DailyRangeAlerts from './widgets/DailyRangeAlerts.vue'
 import CandlestickChart from './widgets/CandlestickChart.vue'
 import TVLiteChart     from './widgets/TVLiteChart.vue'
-import { LINK_COLORS, LINK_COLOR_MAP } from '@/composables/useWidgetBus.js'
+import { LINK_COLORS, LINK_COLOR_MAP } from '@/constants/linkColors.js'
 
 const props = defineProps({
   widgetId:   { type: String,  required: true },
--- a/client/src/components/widgets/CompanyNews.vue
+++ b/client/src/components/widgets/CompanyNews.vue
@@ -200,7 +200,7 @@
  * @emits update-settings    - Settings changed, payload: updated settings object
  */
 import { ref, computed, watch, onMounted, onUnmounted } from 'vue'
-import { useWidgetBus } from '@/composables/useWidgetBus.js'
+import { useDashboardStore } from '@/stores/useDashboardStore.js'
 import { RecycleScroller } from 'vue-virtual-scroller'
 import { useWebSocketClient } from '@/composables/useWebSocketClient.js'
 import { useConfig }          from '@/composables/useConfig.js'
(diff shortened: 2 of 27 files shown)


kuhl-haus-bishop (approved):
CI green, 1438 tests pass. Migration is clean and the test improvements are genuinely better — asserting on store state / DOM rather than mock internals is the right direction.

Two things worth cleaning up before merge:

**1. Inconsistent store mutation in QuoteCoverage.spec.js**

Two tests mutate the store directly rather than going through the action:

```js
// QuoteCoverage.spec.js — direct mutation
store.activeTickers['red'] = 'AAPL'
```

`CompanyNewsCoverage.spec.js` does it correctly:

```js
store.setActiveTicker('blue', 'TSLA')
```

Use `store.setActiveTicker()` throughout. Direct mutation bypasses the action and creates inconsistency across the suite. Tests should exercise the same code path production does.

**2. useWidgetBus.js double-import idiom**

```js
export { LINK_COLORS, LINK_COLOR_MAP } from '@/constants/linkColors.js'
import { LINK_COLORS, LINK_COLOR_MAP } from '@/constants/linkColors.js'
```

Valid ESM, but the `export...from` re-export form is typically for pass-through modules (no local use). When you also need the symbols locally, the conventional form is:

```js
import { LINK_COLORS, LINK_COLOR_MAP } from '@/constants/linkColors.js'
export { LINK_COLORS, LINK_COLOR_MAP }
```

Same semantics, reads cleanly. Cosmetic, but the double-declare will confuse anyone who comes looking.

Everything else is solid. Constants extraction is correct, migration is consistent across all five consumer components, and test isolation via `setActivePinia(createPinia())` is properly in place. refs #281

kuhl-haus-bishop (approved):
Both issues addressed. Ship it.

- `useWidgetBus.js` is now `import` then `export` — clean.
- `QuoteCoverage.spec.js` direct mutations replaced with `store.setActiveTicker()` throughout, consistent with the rest of the suite.
- Bonus: `setActivePinia(createPinia())` added to `EnhancedQuoteV3`, `EnhancedQuoteV3Coverage`, and `EnhancedQuoteV4` specs that were missing it — good catch.
- Circular dependency closed: `useDashboardStore` now imports `LINK_COLORS` from `@/constants/linkColors.js` directly rather than through `useWidgetBus`. That's the right call regardless — stores shouldn't depend on composables for static data.

Two blank lines left behind at the removal site in `NewsFeed.spec.js` — not worth a round-trip. refs #281

oldschool-engineer (approved):

Issue #288, opened 2026-05-08
bug(export): exporting layouts wipes all layouts after browser refresh


kuhl-haus-legion:
## Bug

Clicking **Export All Layouts** (📤) causes all saved layouts to disappear after a browser refresh.

## Root Cause

`exportLayouts` in `DashboardGrid.vue` contained two related defects:

1. **Detached anchor** — the `<a>` element was created but never appended to `document.body` before `a.click()`. In browsers that require the anchor to be connected (notably Safari), clicking a detached anchor with a `blob:` href triggers page navigation instead of a file download. On reload/back-navigation, the Pinia store re-initialises with empty defaults because the in-flight state was never persisted.

2. **Synchronous URL revoke** — `URL.revokeObjectURL(url)` was called synchronously immediately after `a.click()`. If the browser hadn't yet processed the download request, the blob URL was already dead, reinforcing the navigation path.

## Fix

```js
// Before
a.click()
URL.revokeObjectURL(url)

// After
document.body.appendChild(a)
a.click()
document.body.removeChild(a)
setTimeout(() => URL.revokeObjectURL(url), 0)
```

Regression test added: verifies anchor is in DOM when clicked, URL is revoked asynchronously, and store state is unchanged after export.

PR #287, opened 2026-05-08
feat(pinia): consolidate localStorage into stores

kuhl-haus-legion:
refs #282

Chunk 4 of the Pinia migration. Replaces all manual `localStorage` calls for the 6 covered keys with persisted Pinia store fields.

## Changes

- `useWidgetSettingsStore`: populated with 6 fields (`savedLayouts`, `defaultLayoutName`, `isLocked`, `autosaveEnabled`, `maxArticles`, `hasTickersOnly`); `persist: true`
- `DashboardGrid.vue`: 4 localStorage keys migrated to store; `loadFromStorage`/`saveToStorage` become no-ops (plugin handles persistence)
- `NewsFeed.vue`: 2 localStorage keys migrated to store
- Tests updated: store-state assertions replace localStorage spy assertions throughout

Issue #282, opened 2026-05-07
feat(pinia): Chunk 4 — consolidate localStorage into stores


kuhl-haus-legion:
Part of refs #278. Depends on refs #280 (stores must exist).

## Scope

Replace manual `localStorage.getItem/setItem` calls with persisted Pinia store fields using `pinia-plugin-persistedstate`. Consolidates 6 keys across 2 files into a defined schema.

### `DashboardGrid.vue` — 4 keys migrated into `useWidgetSettingsStore`
- Layout storage key (saved layouts JSON)
- Default layout key
- Lock state key
- Autosave enabled key

### `NewsFeed.vue` — 2 keys migrated into `useWidgetSettingsStore`
- Max articles limit
- Has-tickers-only filter flag

### Per file
Remove `localStorage.getItem/setItem` calls; read and write from `useWidgetSettingsStore` fields marked `persist: true`.

## Done when

No manual `localStorage` calls remain for any state covered by the stores. Existing behavior is preserved (values survive page reload). Full test suite passes.

PR #289, opened 2026-05-08
fix(export): append anchor to DOM and revoke URL async to prevent navigation

kuhl-haus-legion:
refs #288

## Problem

Exporting layouts wipes all saved layouts after a browser refresh. The anchor element was detached from the DOM when clicked, causing browsers (especially Safari) to navigate to the blob URL instead of downloading. On reload, the store re-initialises empty. Synchronous `URL.revokeObjectURL` compounded the issue by invalidating the URL before the download request was processed.

## Fix

- Append `<a>` to `document.body` before `a.click()`, remove it immediately after — ensures the download attribute is honoured by all browsers
- Move `URL.revokeObjectURL` into `setTimeout(..., 0)` — defers revocation until after the browser has queued the download

## Tests

Updated `exportLayouts` tests to use a real `HTMLAnchorElement` (not a plain object mock) so that `document.body.appendChild/removeChild` operate correctly. Added assertions for:
- Anchor appended to body before click
- Anchor removed from body after click
- URL revoked asynchronously (not synchronously)
- Store state unchanged after export (regression)

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -499,8 +499,10 @@ const exportLayouts = () => {
   const a = document.createElement('a')
   a.href = url
   a.download = `dashboard-layouts-${Date.now()}.json`
+  document.body.appendChild(a)
   a.click()
-  URL.revokeObjectURL(url)
+  document.body.removeChild(a)
+  setTimeout(() => URL.revokeObjectURL(url), 0)
 }
 
 const importLayouts = (event) => {
--- a/client/src/components/__tests__/DashboardGridOperations.spec.js
+++ b/client/src/components/__tests__/DashboardGridOperations.spec.js
@@ -494,25 +494,41 @@ describe('empty store initial state', () => {
 // ─────────────────────────────────────────────────────────────────────────────
 
 describe('exportLayouts', () => {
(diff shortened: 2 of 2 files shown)


kuhl-haus-bishop (approved):
Fix is correct. Tests are thorough and cover exactly the right things.

The `makeAnchorMock()` helper using a real `HTMLAnchorElement` is the right call — plain object mocks don't participate in `appendChild`/`removeChild`, which is the whole point of the fix. The assertion sequence (append → click → remove → revoke not yet called → flush → revoke called) maps directly to the failure mode being addressed. Regression test is a good add.

One process note: this diff includes the full `useWidgetSettingsStore` Pinia migration from #282. Make sure #287 merges first (or rebase this branch on top of it after merge) to avoid a conflict on that file. refs #288

kuhl-haus-bishop (approved):
Clean. Commit history is tidy, diff is scoped correctly — just the export fix and its tests.

Production change is correct and minimal. Tests cover all the guarantees: anchor appended before click, removed after, URL not revoked synchronously, revoked after the async flush. Regression test pins the store-immutability invariant. `makeAnchorMock()` using a real `HTMLAnchorElement` is the right design.

CI still running on the new push but the prior run on this exact code was green. refs #288

oldschool-engineer (approved):

Issue #292, opened 2026-05-08
feat(alerts): Chunk 2 — AlertSoundPicker component + DailyRangeAlerts alert config


kuhl-haus-legion:
Part of refs #290. Depends on refs #291.

## Deliverables

**`AlertSoundPicker.vue`** (reusable, used in widget settings and Alert Manager):
- Props: `modelValue: string`, `showDefault: boolean`
- Emits: `update:modelValue`
- Renders: `<select>` of available sounds + ▶ preview button
- When `showDefault` is true, prepends a `Default (Sound Name)` option

**DailyRangeAlerts alert config:**
- Settings panel: alert toggle (`settings.alertEnabled`) + `AlertSoundPicker` (`settings.alertSound`)
- Trigger logic:
  - `seenIds` ref (Set) seeded on mount from initial filtered list — no alert on first load
  - Watch filter inputs; on change clear and re-seed `seenIds`, no alert
  - Watch filtered list; on update collect new IDs not in `seenIds`, fire once if any and `alertEnabled`
  - Resolves effective sound: `settings.alertSound ?? widgetSettingsStore.defaultAlertSound`
  - Calls `alertStore.fire(soundId, { timestamp, widgetLabel, widgetType, linkColor, count })`

**Tests:**
- `AlertSoundPicker.spec.js`
- `DailyRangeAlertsCoverage.spec.js` additions

PR #296, opened 2026-05-08
feat(alerts): Chunk 2 — AlertSoundPicker + DailyRangeAlerts alert config

kuhl-haus-legion:
refs #292

## Changes
- `AlertSoundPicker.vue` — reusable sound picker with preview button
- `WidgetWrapper.vue` — passes `userLabel` prop to dynamic child component
- `DailyRangeAlerts.vue` — `userLabel` prop, alert config UI, trigger logic (seenIds + filter-reset)
- `AlertSoundPicker.spec.js` — new test file (8 tests)
- `DailyRangeAlertsCoverage.spec.js` — alert UI and trigger tests added (14 new tests)

All 1472 tests passing.

### DIFF
--- a/client/src/components/AlertSoundPicker.vue
+++ b/client/src/components/AlertSoundPicker.vue
@@ -1,0 +1,38 @@
+<template>
+  <span class="alert-sound-picker">
+    <select
+      :value="modelValue ?? ''"
+      @change="$emit('update:modelValue', $event.target.value || null)"
+      class="sound-select"
+    >
+      <option v-if="showDefault" value="">Default ({{ defaultSoundLabel }})</option>
+      <option v-for="s in ALERT_SOUNDS" :key="s.id" :value="s.id">{{ s.label }}</option>
+    </select>
+    <button type="button" @click="preview" class="preview-btn" title="Preview sound">▶</button>
+  </span>
+</template>
+
+<script setup>
+import { computed } from 'vue'
+import { ALERT_SOUNDS, ALERT_SOUND_MAP } from '@/constants/alertSounds.js'
+import { useWidgetSettingsStore } from '@/stores/useWidgetSettingsStore.js'
+
+const props = defineProps({
+  modelValue:  { type: String,  default: null },
+  showDefault: { type: Boolean, default: false },
+})
+defineEmits(['update:modelValue'])
+
+const widgetSettingsStore = useWidgetSettingsStore()
+
+const defaultSoundLabel = computed(() => {
+  const id = widgetSettingsStore.defaultAlertSound
+  return ALERT_SOUND_MAP[id]?.label ?? id
+})
+
+function preview() {
+  const id = props.modelValue || widgetSettingsStore.defaultAlertSound
+  const sound = ALERT_SOUND_MAP[id]
+  if (sound) new Audio(sound.src).play().catch(() => {})
+}
+</script>
--- a/client/src/components/WidgetWrapper.vue
+++ b/client/src/components/WidgetWrapper.vue
@@ -55,6 +55,7 @@
         :link-color="linkColor"
         :is-mobile="isMobile"
         :settings="settings"
+        :user-label="userLabel"
         @update-col-widths="$emit('update-col-widths', $event)"
         @update-settings="$emit('update-settings', $event)"
       />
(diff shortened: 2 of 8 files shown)


kuhl-haus-bishop (changes requested):
CI green. Production code is correct throughout — trigger logic, filter-reset watcher, sound picker, all solid. Two issues.

**1. Missing filter-reset test (blocking)**

The filter-reset watcher is the implementation of the N1 design decision. There is no test for it. A test that exercises filter input changes and verifies pre-existing-but-newly-visible events do *not* trigger `alertStore.fire` needs to exist. Without it the watcher could be removed and the suite stays green.

**2. Dual Pinia instances in `AlertSoundPicker.spec.js` (non-blocking but fix it now)**

`beforeEach` calls `setActivePinia(createPinia())` then `mountPicker` passes `global: { plugins: [createPinia()] }` — a different instance. VTU installs its own so the active-pinia instance is shadowed inside the component. Tests pass today because none set store state before mounting, but this is a footgun. Drop the `createPinia()` from `global.plugins`:

```js
function mountPicker(props = {}) {
  return mount(AlertSoundPicker, { props })
}
```

refs #292

kuhl-haus-bishop (approved):
Both issues resolved. Filter-reset test is exactly right — seeds with a tight filter, loosens it, verifies no false-positive alert. Dual Pinia footgun is gone from AlertSoundPicker.spec.js.

Bonus: the filter-reset watcher was improved to watch `config.value` fields directly rather than local refs, with a comment explaining the watcher-ordering guarantee. That's cleaner than what was there before.

refs #292

kuhl-haus-bishop (approved):
Clean. All four files in the fix commit are correct.

The `vite.config.js` changes are necessary and well-reasoned — `base: './'` is required because the Dockerfile copies `dist/assets/*` flat into `static/dist/`, so absolute `/assets/...` URLs would 404 behind py4web's `/static/` prefix. Relative URLs resolve correctly from the JS file's location. The content hash in the filename pattern is a good addition regardless. This is the right PR for it — first chunk that introduces audio file imports.

refs #292

kuhl-haus-bishop (changes requested):
The `preloadAll()` implementation is correct — caching strategy, `currentTime = 0`, `typeof a.load` guard, integration point. One test to fix before merge.

**`useAlertStore.spec.js` — second `preloadAll` test has a no-op assertion**

```js
test('with preloadAll called twice expect Audio constructed only once per sound', () => {
  // ...
  expect(store).toBeDefined()  // verifies nothing
})
```

The test name promises deduplication behaviour. The assertion delivers nothing. The comment says the jsdom stub lacks `.load()` so the cache never fills — but that's fixable. Add `.load` to the stub:

```js
const AudioMock = vi.fn(function () {
  return { play: playImpl, load: vi.fn() }
})
```

Now `_audioCache` populates on the first call, the second call skips construction, and the assertion becomes `expect(AudioMock).not.toHaveBeenCalled()` after clearing. That actually verifies what the test name claims.

refs #292

kuhl-haus-bishop (approved):
Fixed. `AudioMock` now has `.load`, cache fills on first call, second call assertion is real.

Two minor notes, not worth a round-trip:
1. First `preloadAll` test comment still says "cache stays empty" — stale now that the stub has `.load()`.
2. `_audioCache` is module-level, not cleared by `beforeEach` — test 2's "first call" actually hits a cache already filled by test 1. Assertion still passes correctly but for the wrong reason when run in sequence. Would need a test-only `clearCache()` export or per-instance Map to fix cleanly.

refs #292

kuhl-haus-bishop (changes requested):
Walking back my approval — the two notes I flagged should be fixed here, not deferred.

**1. Stale comment in first `preloadAll` test**

```js
// in tests the stub returns a plain object without .load() so the cache stays empty
```

The stub now has `.load: vi.fn()` so the cache does fill. Update the comment to reflect reality.

**2. `_audioCache` module-level Map not cleared between tests**

`_audioCache` persists across tests. Test 1 fills it; test 2's "first call" hits an already-populated cache and constructs 0 Audio objects instead of 14. The assertion still passes but for the wrong reason when run in sequence.

Fix: export a `_clearAudioCacheForTesting` (or similar) from the store module and call it in `beforeEach`, or move `_audioCache` into the store instance as a `ref(new Map())` so `setActivePinia(createPinia())` resets it naturally with each fresh store.

refs #292

kuhl-haus-bishop (approved):
Both issues fixed correctly. Stale comment updated. `_audioCache` moved into the store function — per-instance now, resets with `setActivePinia(createPinia())`, test isolation is clean. refs #292

oldschool-engineer (approved):

Issue #293, opened 2026-05-08
feat(alerts): Chunk 3 — NewsFeed alert config


kuhl-haus-legion:
Part of refs #290. Depends on refs #291.

## Deliverables

**NewsFeed alert config** (same pattern as Chunk 2):
- Settings panel: alert toggle (`settings.alertEnabled`) + `AlertSoundPicker` (`settings.alertSound`)
- Trigger logic:
  - `seenIds` ref (Set) seeded on mount from initial filtered list — no alert on first load
  - Watch filter inputs; on change clear and re-seed `seenIds`, no alert
  - Watch filtered list; on update collect new IDs not in `seenIds`, fire once if any and `alertEnabled`
  - Resolves effective sound: `settings.alertSound ?? widgetSettingsStore.defaultAlertSound`
  - Calls `alertStore.fire(soundId, { timestamp, widgetLabel, widgetType, linkColor, count })`

**Tests:**
- `NewsFeedCoverage.spec.js` additions

PR #297, opened 2026-05-08
feat(alerts): Chunk 3 — NewsFeed alert config

kuhl-haus-legion:
refs #293

## Changes
- `AlertSoundPicker.vue` — reusable sound picker with preview (also in Chunk 2; identical spec)
- `WidgetWrapper.vue` — passes `userLabel` prop to dynamic child component
- `NewsFeed.vue` — `userLabel` prop, alert toggle + sound picker in controls bar, trigger logic (seenIds + filter-reset)
- `NewsFeedCoverage.spec.js` — alert UI and trigger tests added

### DIFF
--- a/client/src/components/widgets/NewsFeed.vue
+++ b/client/src/components/widgets/NewsFeed.vue
@@ -47,6 +47,22 @@
       <select v-model.number="maxArticles" class="max-articles-select" title="Max articles loaded">
         <option v-for="n in MAX_ARTICLES_OPTIONS" :key="n" :value="n">{{ n }}</option>
       </select>
+
+      <!-- Audio alert toggle -->
+      <button
+        :class="['filter-btn', config_alertEnabled ? 'filter-btn--active' : '']"
+        :title="config_alertEnabled ? 'Audio alerts ON \u2014 click to disable' : 'Audio alerts OFF \u2014 click to enable'"
+        @click="onAlertEnabledChange(!config_alertEnabled)"
+        data-testid="alert-toggle"
+      >&#x1F514;</button>
+
+      <!-- Sound picker (shown when alerts enabled) -->
+      <AlertSoundPicker
+        v-if="config_alertEnabled"
+        :model-value="config_alertSound"
+        :show-default="true"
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Clean. Production code is correct and the tests cover what matters.

`alertReady` flag is the right call for NewsFeed specifically — the nextTick seed window needs a gate, and the initial-load test validates it. Filter-reset watcher has the right three inputs (`hasTickersOnly`, `activeTicker`, `searchQuery` — the only ones that drive `filteredNews`; `maxArticles` is applied at ingestion not display so it correctly stays out). Filter-reset test is present and covers the key scenario.

One cosmetic note: `config_alertEnabled` / `config_alertSound` use underscores in otherwise camelCase names. Not blocking, but worth normalising to `configAlertEnabled` / `configAlertSound` for consistency with the rest of the component.

refs #293

kuhl-haus-bishop (approved):
Fixed. Consistent camelCase throughout. Nothing else to flag. refs #293

kuhl-haus-bishop (approved):
New commit is clean — camelCase rename correct throughout, Audio stub in NewsFeedCoverage matches DRA's pattern, vite.config matches PR 296.

One follow-up item (not blocking this PR): `onAlertEnabledChange` in NewsFeed doesn't call `alertStore.preloadAll()` the way DRA does. The `fire()` fallback still works, but if NewsFeed is the only alert-enabled widget, sounds won't be pre-cached. Parallel branch timing means `preloadAll()` wasn't available when this was written — worth a follow-up patch after these chunks merge.

refs #293

kuhl-haus-bishop (approved):
Both issues resolved — camelCase throughout, `preloadAll()` called on enable. Clean. refs #293

kuhl-haus-bishop (approved):
Clean. Correctly scoped now — shared files (AlertSoundPicker, WidgetWrapper, vite.config) removed from this branch since they land via #296 first. PR 297 is purely the NewsFeed delta. Nothing new to flag. refs #293

oldschool-engineer (approved):

PR #299, opened 2026-05-11
fix(charts): ticker bus persistence, commit-gated fetch, and bidirectional bus for TVLiteChart and CandlestickChart

kuhl-haus-legion:
### Bug 1 — Ticker lost on touch/scroll (mobile layout re-render)

When the widget bus sets a ticker, `headerTickerInput` was updated but `emitSettings()` was not called — so the ticker was never persisted to `props.settings`.

`vue3-grid-layout-next` creates new item objects during touch/scroll compaction for any widget that needs repositioning. The new object reference triggers the settings watcher, which resets `headerTickerInput` to the settings value — which is empty because the ticker was never persisted. Widgets at `y=0` (the first chart added) are unaffected because they never need repositioning and their settings reference never changes. Subsequent charts at `y>0` all lose their tickers.

**Fix:** call `emitSettings()` in the bus watcher so the ticker is immediately persisted to settings on every bus update.

### Bug 2 — Premature fetch on keypress

`tickerLocal` was `computed(() => headerTickerInput.value...)`, so every `@input` keystroke changed `tickerLocal`, which triggered `watch(tickerLocal)` and fired `fetchBars()`. Charts were fetching live API data for every partial ticker typed ('A', 'AP', 'APP', 'APPL', 'AAPL').

**Fix:** `tickerLocal` is now a `ref`, initialized from settings. It is only updated on explicit commit (Go click, Enter key, bus update, settings re-apply). Keystrokes update `headerTickerInput` only — the watcher never fires mid-typing.

### Bug 3 — One-way bus (typed tickers not broadcast to linked widgets)

`onGoTicker()` called `emitSettings()` but never `store.setActiveTicker()`. Ticker events flowed INTO charts from the bus (scanner → chart) but never OUT (chart → other linked charts/widgets). Typing a ticker in one chart and clicking Go had no effect on sibling charts sharing the same link color.

**Fix:** `onGoTicker()` now calls `dashboardStore.setActiveTicker(linkColor, sym)` after committing, making chart widgets first-class bus participants in both directions.

## Implementation notes

- `tickerLocal computed` → `ref`: the bus watcher and settings watcher both update `tickerLocal` explicitly; the `watch(tickerLocal, { immediate: true })` watcher only fires on those committed updates, not on keystrokes.
- Bus watcher is guarded with `t !== tickerLocal.value` to prevent double `emitSettings()` on Go/Enter: `onGoTicker()` broadcasts to the store → store change echoes back through `activeTicker` → watcher fires → guard detects `tickerLocal` is already set → skips the redundant emit. External bus updates (scanner row clicks) still emit because the ticker is genuinely new.
- All fixes applied to both `TVLiteChart` and `CandlestickChart`.

## Tests

10 new regression tests total (5 per widget), all were red before the fix:

**TVLiteChart.spec.js:**
1. `bus activeTicker emits update-settings with the ticker (persistence)`
2. `ticker set via bus survives settings prop re-application (simulates touch/scroll layout re-render)`
3. `typing in header input does NOT trigger fetch before committing`
4. `Go button broadcasts committed ticker to linked bus color`
5. `Enter key in header input broadcasts ticker to linked bus color`

**CandlestickChart.spec.js:**
6. `bus activeTicker emits update-settings with the ticker (persistence)`
7. `ticker set via bus survives settings prop re-application (simulates touch/scroll layout re-render)`
8. `typing in header input does NOT trigger fetch before committing`
9. `Go button broadcasts committed ticker to linked bus color`
10. `Enter key in header input broadcasts ticker to linked bus color`

Full suite: 49 test files, 1524 tests — all pass.

### DIFF
--- a/client/src/components/widgets/CandlestickChart.vue
+++ b/client/src/components/widgets/CandlestickChart.vue
@@ -246,8 +246,12 @@ const tickerLocal = ref(config.value.ticker?.trim().toUpperCase() || null)
 // vue3-grid-layout-next does on touch/scroll events for widgets that need
 // repositioning (typically any chart after the first one at y=0).
 watch(activeTicker, (t) => {
-  if (t) {
-    headerTickerInput.value = t
+  if (!t) return
+  headerTickerInput.value = t
+  // Guard against self-echo: onGoTicker() calls setActiveTicker() which bounces
+  // back through this watcher. If tickerLocal already equals t, settings were
+  // already emitted by onGoTicker — skip to avoid a double update-settings event.
+  if (t !== tickerLocal.value) {
     tickerLocal.value = t
     emitSettings()
   }
--- a/client/src/components/widgets/CandlestickChart.vue
+++ b/client/src/components/widgets/CandlestickChart.vue
@@ -233,9 +233,16 @@ const showSettings        = ref(false)
 // ── Resolved ticker ───────────────────────────────────────────────────────────
 const tickerLocal = computed(() => headerTickerInput.value?.trim().toUpperCase() || null)
 
-// Bus auto-fills the header input when it fires
+// Bus auto-fills the header input and persists the ticker to settings.
+// Without emitSettings() here, the ticker lives only in headerTickerInput and is
+// lost whenever props.settings is re-applied with a new object reference — which
+// vue3-grid-layout-next does on touch/scroll events for widgets that need
+// repositioning (typically any chart after the first one at y=0).
 watch(activeTicker, (t) => {
-  if (t) headerTickerInput.value = t
+  if (t) {
+    headerTickerInput.value = t
+    emitSettings()
+  }
 })
 
 const onGoTicker = () => {
--- a/client/src/components/widgets/CandlestickChart.vue
+++ b/client/src/components/widgets/CandlestickChart.vue
@@ -137,6 +137,7 @@ import {
 import VChart from 'vue-echarts'
 import { useScannerLink } from '@/composables/useScannerLink.js'
 import { useConfig } from '@/composables/useConfig.js'
+import { useDashboardStore } from '@/stores/useDashboardStore.js'
 import {
   calcEMA, calcSMA, calcVWMA, calcVWAP, calcMACD, calcVolumeAvg,
 } from '@/utils/chartIndicators.js'
@@ -204,8 +205,11 @@ const config = computed(() => ({ ...DEFAULT_SETTINGS, ...props.settings }))
 // ── Config (massiveApiKey) ───────────────────────────────────────────────────
 const { config: appConfig } = useConfig()
 
+// ── Dashboard store (for bidirectional bus writes)
+const dashboardStore = useDashboardStore()
+
 // ── Scanner bus ───────────────────────────────────────────────────────────────
-// CandlestickChart is read-only from the bus — no onRowClick needed
+// CandlestickChart now bidirectional: reads from bus (activeTicker), writes to bus (onGoTicker)
 const { activeTicker } = useScannerLink(computed(() => props.linkColor))
 
 // ── Local state ───────────────────────────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


kuhl-haus-legion:
## Additional fixes pushed (same branch)

Two more related bugs found and fixed in this commit:

### Bug 3 — Premature fetch on keypress
`tickerLocal` was `computed(() => headerTickerInput.value...)`, so every `@input` keystroke changed `tickerLocal`, which triggered `watch(tickerLocal)` and fired `fetchBars()`. Charts were fetching live data for every partial ticker typed ('A', 'AP', 'APP', 'APPL', 'AAPL').

**Fix:** `tickerLocal` is now a `ref` only updated on explicit commit (Go/Enter/bus/settings). Keystrokes update `headerTickerInput` only — the watcher never fires mid-typing.

### Bug 4 — One-way bus (typed tickers not broadcast to linked widgets)
`onGoTicker()` called `emitSettings()` but never `store.setActiveTicker()`. Ticker events flowed INTO charts from the bus (scanner → chart) but never OUT (chart → other linked charts). Typing a ticker in one chart had no effect on sibling charts sharing the same link color.

**Fix:** `onGoTicker()` now calls `dashboardStore.setActiveTicker(linkColor, sym)` so the committed ticker is broadcast to all linked widgets.

Both fixes applied to `TVLiteChart` and `CandlestickChart`.

3 new regression tests (all were red before fix):
1. `typing in header input does NOT trigger fetch before committing`
2. `Go button broadcasts committed ticker to linked bus color`
3. `Enter key in header input broadcasts ticker to linked bus color`

Full suite: 49 files, 1519 tests — all pass.

kuhl-haus-bishop (changes requested):
## Changes Requested

CI is green and the root cause analysis is correct — the `vue3-grid-layout-next` object-reference churn explains the bug precisely. Fix logic is sound. Three things to fix before this merges.

---

### 🚫 No regression tests for `CandlestickChart`

The same three bugs were fixed in `CandlestickChart.vue` — bus persistence, keystroke-triggered fetches, bidirectional bus broadcast — but `CandlestickChart.spec.js` got zero new tests. The existing test at line 606 (`bus activeTicker updates the header ticker input`) only asserts that the input field updates; it does not assert that `update-settings` is emitted with the ticker, does not simulate settings re-application, and does not test Go/Enter bus broadcast.

If someone reverts the `CandlestickChart` side of this fix, nothing catches it. The five new tests in `TVLiteChart.spec.js` need counterparts in `CandlestickChart.spec.js`.

---

### 🚫 Double `emitSettings()` on Go/Enter when `linkColor` is set

Trace through `onGoTicker()`:

```js
if (props.linkColor && sym) dashboardStore.setActiveTicker(props.linkColor, sym)
emitSettings()  // emit #1
```

`setActiveTicker()` updates the store → `activeTicker` ref changes → Vue schedules the watcher → next tick:

```js
watch(activeTicker, (t) => {
  ...
  emitSettings()  // emit #2
})
```

The parent receives two identical `update-settings` events on every Go/Enter with a linked color. Not a correctness bug, but a guaranteed double re-render/re-save on every committed ticker entry, and it will be confusing to debug later. Fix: guard the watcher against the self-triggered case, or skip the explicit `emitSettings()` in `onGoTicker()` when a store write will cause the watcher to emit anyway.

---

### 🚫 PR description underscopes the change

The description says "two regression tests added" — five were added. More importantly:

- The `tickerLocal` computed→ref change is a real behavior fix (prevents keystroke-triggered fetches) and deserves its own bullet, not burial as an implementation detail
- The bidirectionality change (`onGoTicker` now writing to the bus) is mentioned only in a code comment — it belongs in the description

These are substantive behavioral changes. Keep the description accurate for future git archaeology.

kuhl-haus-legion:
All three review items addressed — pushed as a single follow-up commit.

**CandlestickChart tests:** 5 new tests added to `CandlestickChart.spec.js` mirroring all 5 TVLiteChart tests (bus persistence × 2, commit-gated fetch × 1, bidirectional bus × 2). All were red before the fix.

**Double emitSettings():** Guarded the bus watcher with `t !== tickerLocal.value`. When `onGoTicker()` sets `tickerLocal` then broadcasts to the store, the echo back through the watcher sees the same ticker and skips the redundant emit. External bus updates (scanner row clicks, sibling charts) still emit because the ticker is genuinely new to `tickerLocal`.

**PR description:** Updated to accurately describe all 3 bugs, the `tickerLocal computed→ref` behavioral change, the double-emit guard, and all 10 tests.

Suite: 49 files, 1524 tests — all pass.

kuhl-haus-bishop (approved):
All three issues addressed. Approving.

**CandlestickChart tests** — all five regression tests ported over, symmetric coverage with TVLiteChart. ✓

**Double emitSettings()** — the `t !== tickerLocal.value` guard is the right fix. `onGoTicker()` sets `tickerLocal` before the store write, so the self-echo is cleanly suppressed. External bus updates (scanner → chart) still emit because the ticker is genuinely new. ✓

**PR description** — completely rewritten. Three bugs, three fixes, 10 tests accounted for individually. This is what good description hygiene looks like. ✓

CI green across the board. LGTM.

oldschool-engineer (approved):

PR #300, opened 2026-05-11
fix(autosave): preserve description and created timestamp on autosave

kuhl-haus-legion:
## Bug

`autoSaveLayout()` replaced the entire layout object with a fresh one on every write, silently dropping two fields:

- `description` — the user-provided label set in the Save dialog
- `created` — the original creation timestamp

The first autosave after any widget change while the layout is unlocked nuked both. Layout descriptions would appear populated after a manual save, then vanish the next time the user dragged a widget.

```js
// Before — fresh object, description and created lost
savedLayouts.value[name || AUTOSAVE_KEY] = {
  layout: layoutCopy,
  widgetCounter: widgetCounter,
  dashboardColNum: dashboardColNum.value,
  modified: Date.now(),
}

// After — read existing entry first, carry forward description and created
const existing = savedLayouts.value[name || AUTOSAVE_KEY]
savedLayouts.value[name || AUTOSAVE_KEY] = {
  layout: layoutCopy,
  widgetCounter: widgetCounter,
  dashboardColNum: dashboardColNum.value,
  created:     existing?.created     ?? Date.now(),
  description: existing?.description ?? '',
  modified:    Date.now(),
}
```

`saveLayout()` already handled this correctly via `existing?.created` — `autoSaveLayout()` just never got the same treatment.

The `__autosave__` fallback key (used when no named layout is active) is also covered: `existing` is undefined on first write, so `created` defaults to `Date.now()` and `description` to `''`, which is correct.

## Test

One new regression test in `DashboardGridOperations.spec.js`:

> `with existing layout having description and created expect autosave preserves both`

Was red before fix, green after. Full suite: 49 files, 1525 tests — all pass.

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -514,11 +514,14 @@ const autoSaveLayout = () => {
   autoSaveTimeout = setTimeout(() => {
     const layoutCopy = JSON.parse(JSON.stringify(layout.value))
     const name = selectedLayoutName.value
+    const existing = savedLayouts.value[name || AUTOSAVE_KEY]
     savedLayouts.value[name || AUTOSAVE_KEY] = {
       layout: layoutCopy,
       widgetCounter: widgetCounter,
       dashboardColNum: dashboardColNum.value,
-      modified: Date.now()
+      created:     existing?.created     ?? Date.now(),
+      description: existing?.description ?? '',
+      modified:    Date.now(),
     }
 
     isAutoSaving.value = false
--- a/client/src/components/__tests__/DashboardGridOperations.spec.js
+++ b/client/src/components/__tests__/DashboardGridOperations.spec.js
@@ -147,6 +147,7 @@ beforeEach(() => {
 
 afterEach(() => {
   vi.restoreAllMocks()
+  vi.useRealTimers() // ensure fake timers never leak between tests regardless of assertion failures
 })
 
 // ─────────────────────────────────────────────────────────────────────────────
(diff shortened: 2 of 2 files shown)


kuhl-haus-bishop (approved):
Clean fix. Approving.

The bug is real and the root cause is well-explained — `saveLayout()` already did this correctly, `autoSaveLayout()` just never got the same treatment. The fix is minimal and exact: read `existing` first, carry `created` and `description` forward with `??` defaults. Nothing surprising in the diff.

**Test:** Naming convention, AAA, fake timers used correctly, asserts on the observable state (`savedLayouts`) rather than internals. One note: if the assertion throws before `vi.useRealTimers()`, fake timers leak into downstream tests. Not blocking given it is likely the prevailing pattern in this suite, but a `try/finally` or `afterEach` cleanup is the safer long-term pattern.

**Coverage gap (not blocking):** There is no regression test for the `__autosave__` key preserving `created` across subsequent writes. The code path is identical and the fix handles it correctly, but the symmetry would be nice. Worth a follow-up if the test suite gets a coverage pass.

CI green across the board. LGTM.

kuhl-haus-bishop (changes requested):
Superseding my approval — two issues I soft-pedaled should be blocking.

---

### 🚫 Fake timer leak on assertion failure

`vi.useFakeTimers()` is called in the test body with no `try/finally` or `afterEach` cleanup. If the assertion throws, `vi.useRealTimers()` never runs and fake timers leak into every downstream test in the suite. Fix: move `vi.useFakeTimers()` / `vi.useRealTimers()` into `beforeEach` / `afterEach`, or wrap the test body in `try/finally`.

---

### 🚫 No regression test for `__autosave__` key preserving `created`

The fix covers both named layouts and the `__autosave__` fallback key. The named layout case is tested. The `__autosave__` case — where an existing autosave entry already has a `created` timestamp and it must survive subsequent autosaves — is not. The code path is identical, but the regression coverage is asymmetric. Add a second test: mount with no named layout selected, trigger two autosaves, assert `created` on the second write matches the first.

kuhl-haus-legion:
Both blocking items addressed, force-pushed as an amended commit.

**Fake timer leak:** Added `vi.useRealTimers()` to the suite-level `afterEach`. Covers all fake-timer tests in this file — including the two existing ones — regardless of whether assertions pass or fail. The inline `vi.useRealTimers()` calls in individual tests remain and are now redundant but harmless.

**`__autosave__` regression test:** Mounts with no named layout selected (`selectedLayoutName` starts as `''`), triggers two autosaves with `vi.advanceTimersByTime(60_000)` between them so `Date.now()` would return a different value on the second write. Asserts `created` on the second write equals `created` from the first write. Passes with the fix, would fail on the pre-fix code.

Suite: 49 files, 1526 tests — all pass.

kuhl-haus-bishop (approved):
All issues addressed. Approving.

`vi.useRealTimers()` in `afterEach` is the right fix — unconditional cleanup regardless of assertion failures, and it covers every test in the file, not just this one.

The `__autosave__` test is solid. Advancing fake time 60s between writes means a broken implementation would produce a detectably different `created` — this is a real regression test, not a coincidence pass. CI green. LGTM.

oldschool-engineer (approved):

Issue #278, opened 2026-05-07
feat(pinia): adopt Pinia for UI state management


kuhl-haus-legion:
## Summary

The trading dashboard currently manages shared UI state through `useWidgetBus` — a Vue `reactive()` module singleton — and ad-hoc `localStorage` keys scattered across individual widgets. This works but has a real scaling ceiling: no DevTools integration, no persistence story, and a subscription tax on every new widget that needs shared state.

This issue tracks adopting Pinia as the canonical layer for UI state.

**Note:** Real-time market data (quotes, OHLCV, alerts) is delivered via WebSocket directly to widgets through `useWebSocketClient`. It does not flow through `useWidgetBus` and is not in scope here.

---

## What Changes

- **Active ticker** (per link color) moves from `useWidgetBus` into `useDashboardStore`
- **Filter state** (DailyRangeAlerts) moves into `useDashboardStore`
- **Widget settings** (scattered `localStorage` keys) consolidated into `useWidgetSettingsStore` with `pinia-plugin-persistedstate`
- **`useWidgetBus`** is retained only for news freshness timestamps and flame icons — no Pinia equivalent needed there

---

## Definition of Done

- `useWidgetBus` is used only for news freshness timestamps and flame icons
- Active ticker, filter state, widget settings, and layout are all in Pinia stores
- All 22 affected test files use direct store state (`setActivePinia(createPinia())`)
- No manual `localStorage.getItem/setItem` calls remain for state covered by stores

---

## Why Pinia Over DIY `reactive()`

A `reactive()` singleton gives you reactivity but none of the infrastructure. Pinia adds:
- **Named, typed stores** with a canonical API
- **Vue DevTools integration** — inspect current values, action history, component consumers
- **Persistence via `pinia-plugin-persistedstate`** — replaces ad-hoc `localStorage` calls
- **Test ergonomics** — `setActivePinia(createPinia())` replaces bus mock/setup boilerplate

---

## Work Breakdown

| Chunk | Size | Scope |
|---|---|---|
| Chunk 1 — Install + wire up | XS | 1 `npm install`, ~10 lines in `main.js` |
| Chunk 2 — Create stores + migrate tests | S | 2 new store files, 22 test files updated |
| Chunk 3 — Migrate active ticker consumers | S | 1 composable + 8 widget files |
| Chunk 4 — localStorage migration | S | 6 `localStorage` keys across 2 files |

Sub-issues track each chunk individually.

---

## Risk

[...]

PR #301, opened 2026-05-13
docs: add Development section covering Vue DevTools and Pinia state inspection

kuhl-haus-legion:
Adds a **Development** section to the README covering Vue DevTools setup and Pinia state inspection.

## What's included

- How to enable devtools (`npm run dev` vs `vite build --mode development`)
- How Vue/Vite detects production (`__DEV__` flag + tree-shaking)
- Observing mutation history in the DevTools Timeline tab
- Programmatic observation with `$subscribe` (including the dev-only `mutation.events` caveat)
- Tracing which action caused a state change with `$onAction`
- Verifying devtools are active in the browser console
- `pinia-plugin-persistedstate` tip for confirming localStorage writes

refs #278

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,76 @@ The platform consists of four main packages:
 - **Frontend Application** ([kuhl-haus-mdp-app](https://github.com/kuhl-haus/kuhl-haus-mdp-app)) — Web-based user interface and API *(this repo)*
 - **Deployment Automation** ([kuhl-haus-mdp-deployment](https://github.com/kuhl-haus/kuhl-haus-mdp-deployment)) — Docker Compose, Ansible playbooks and Kubernetes manifests for environment provisioning
 
+## Development
+
+### Pinia State Inspection with Vue DevTools
+
+Install the [Vue DevTools browser extension](https://devtools.vuejs.org/) (Chrome/Edge/Firefox).
+DevTools hooks are only active in dev builds — they are compiled out of production bundles entirely.
+
+**Start the dev server** (devtools enabled automatically):
+
+```bash
+npm run dev
+```
+
+To produce a built artifact with devtools intact (e.g. for staging):
+
+```bash
+vite build --mode development
+```
+
+#### How Vue knows it's in production
+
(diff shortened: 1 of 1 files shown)


kuhl-haus-bishop (commented):
## Bishop's Review ♟️

CI green. Most of the technical content is accurate — `$subscribe` API, `$onAction` API, `__DEV__` tree-shaking mechanics, `vite build --mode development`, and `window.__VUE_DEVTOOLS_GLOBAL_HOOK__` all check out against the Pinia docs and Vite docs.

### Blocker — `pinia-plugin-persistedstate` tip is factually wrong

> State changes that trigger a localStorage write show up in the Timeline as a `patch object` mutation.

This is incorrect. The mutation type (`'direct'` | `'patch object'` | `'patch function'`) describes *how the state was changed*, not whether the persistence plugin fired. Those types map to:
- `'direct'` — direct property assignment (`store.foo = bar`)
- `'patch object'` — `store.$patch({ foo: bar })`
- `'patch function'` — `store.$patch(state => { state.foo = bar })`

`pinia-plugin-persistedstate` works via `$subscribe` and writes to localStorage after **any** mutation regardless of type. A persisted store using direct mutations in its actions would show `'direct'` in the Timeline, not `'patch object'`. There is no mutation type specific to persistence writes.

**Fix:** Replace the tip with something accurate:

> Every state mutation on a persisted store triggers a localStorage write — `pinia-plugin-persistedstate` subscribes to all mutations via `$subscribe`. In the Timeline, you'll see a `Pinia 🍍` event for each mutation; to confirm the write actually landed, check Application → Local Storage for the `widgetSettings` key after triggering an action.

### Minor — Timeline description overstates action attribution

> Every state change emits a Pinia 🍍 event showing the store id, the action that triggered it...

Direct mutations (property assignments outside of a defined action) don't carry action attribution in the Timeline. Worth a small qualification: *"...the action that triggered it, if applicable..."*

---

Fix the persistedstate tip and this is good to go.

kuhl-haus-bishop (approved):
Both issues addressed correctly. ♟️

oldschool-engineer (approved):

PR #302, opened 2026-06-27
fix(mobile): chart height collapse, resize handle touch target, autosave button visibility

kuhl-haus-legion:
### 1. Charts not rendering on mobile (root cause fixed)

`widget-content` had `overflow: hidden` on mobile. This collapsed the flex chain before browser layout paint, so chart widgets called `clientHeight` at mount time and got `0`. Both TVLiteChart and CandlestickChart rendered invisibly — TVLiteChart's `|| 300` fallback from the previous commit only covered the `createChart()` call, not the container's actual painted height.

**Fix:** Replace `overflow: hidden` + `flex: 1` with `height: 388px` (420px wrapper − 32px header) + `overflow: auto` + `flex: none`. Both chart widgets use `height: 100%` → `flex: 1; min-height: 0` → chart container internally, which now resolves against a real non-zero pixel height.

### 2. Mobile header: all controls visible, version not clipped

The header is a flat `flex` row with no `flex-wrap`. On a phone (375–430px wide), title + status badge + mobile toolbar + version badge all fight for one row. Controls wrap but overflow invisibly; version hangs off-screen.

**Fix:** Enable `flex-wrap` on `.dashboard-header` for mobile. The toolbar (`.layout-controls--mobile`) becomes a full-width second row via `width: 100%`. It's `overflow-x: auto` with hidden scrollbar, so all buttons (layout dropdown, 💾, 🔄/⏸, 🔒, add widget) are reachable by horizontal swipe. Version badge stays in the top row via `margin-left: auto`.

### 3. Resize handle (Bishop's blocking comment, already merged)

Wrapped `.vue-resizable-handle` override in `@media (hover: none) and (pointer: coarse)` so the 44×44px touch target applies only on touchscreens.

### 4. Button touch targets

All `.btn-icon` on mobile get `min-width: 44px; min-height: 44px` (Apple HIG minimum).

---

All 1526 tests pass.

### DIFF
--- a/client/src/assets/base.css
+++ b/client/src/assets/base.css
@@ -57,11 +57,13 @@
 }
 
 /* Enlarge vue3-grid-layout-next resize handle for touch devices (Apple HIG: 44×44pt min) */
-.vue-grid-item > .vue-resizable-handle {
-  width: 44px;
-  height: 44px;
-  background-size: 14px 14px;
-  background-position: bottom 6px right 6px;
+@media (hover: none) and (pointer: coarse) {
+  .vue-grid-item > .vue-resizable-handle {
+    width: 44px;
+    height: 44px;
+    background-size: 14px 14px;
+    background-position: bottom 6px right 6px;
+  }
 }
 
 body {
--- a/client/src/assets/base.css
+++ b/client/src/assets/base.css
@@ -56,6 +56,14 @@
   font-weight: normal;
 }
 
+/* Enlarge vue3-grid-layout-next resize handle for touch devices (Apple HIG: 44×44pt min) */
+.vue-grid-item > .vue-resizable-handle {
+  width: 44px;
+  height: 44px;
+  background-size: 14px 14px;
+  background-position: bottom 6px right 6px;
+}
+
 body {
   min-height: 100vh;
   color: var(--color-text);
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -35,6 +35,11 @@
       </div>
       <button @click="showSaveDialog = true" class="btn-icon" title="Save Layout">💾</button>
       <button
+          @click="autosaveEnabled = !autosaveEnabled"
+          :class="['btn-icon', autosaveEnabled ? '' : 'btn-icon--inactive']"
+          :title="autosaveEnabled ? 'Autosave ON' : 'Autosave OFF'"
+      >{{ autosaveEnabled ? '🔄' : '⏸' }}</button>
+      <button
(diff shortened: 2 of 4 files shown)


client/src/assets/base.css
```diff
@@ -56,6 +56,14 @@
   font-weight: normal;
 }
 
+/* Enlarge vue3-grid-layout-next resize handle for touch devices (Apple HIG: 44×44pt min) */
+.vue-grid-item > .vue-resizable-handle {
+  width: 44px;
+  height: 44px;
+  background-size: 14px 14px;
```
kuhl-haus-bishop: This override has no media query, so it applies on desktop too. iPad (≥ 640px) uses the `GridLayout` path — correct — but so does every mouse-driven desktop session. A 44×44px handle in the bottom-right corner of every widget on desktop is noticeably large and will overlap chart content (time axis, legend).

Scope it to touch-primary devices:

```css
@media (hover: none) and (pointer: coarse) {
  .vue-grid-item > .vue-resizable-handle {
    width: 44px;
    height: 44px;
    background-size: 14px 14px;
    background-position: bottom 6px right 6px;
  }
}
```

`(hover: none) and (pointer: coarse)` matches touchscreens (phone, tablet) and excludes mouse/trackpad users.

kuhl-haus-bishop (changes requested):
Three of the four changes are solid. The `touch-action: pan-y` fix is correct, the flex chain fix for chart height collapse is correct, the `|| 300` fallback is fine (ResizeObserver will correct it after layout paint), and the autosave button addition to the mobile toolbar is the right fix for that omission. All CI green.

One blocking item — see inline comment on `base.css`.

Two non-blocking observations:

**Fixed `height: 420px` on all mobile widgets** — this fixes charts but makes every widget type (TopVolume, NewsFeed, etc.) the same height regardless of content. Fine for now since it's not worse than the previous collapse-to-zero behavior, but worth tracking: non-chart widgets may have excess whitespace or scroll awkwardly. Something to revisit when mobile layout gets a deeper pass.

**Autosave button title text** — mobile uses `'Autosave ON'` / `'Autosave OFF'` while desktop has `'Autosave ON — click to disable'` / `'Autosave OFF — click to enable'`. Title attributes have limited value on mobile (no hover), so shorter is arguably fine — just note the inconsistency if you ever unify them.

kuhl-haus-bishop (approved):
Blocker addressed — resize handle is now correctly scoped to `(hover: none) and (pointer: coarse)`. Desktop unaffected. All CI green. Good to merge.

oldschool-engineer (approved):

PR #303, opened 2026-06-27
fix(mobile): rework header layout and fix chart height chain

kuhl-haus-legion:
### 1. Charts not rendering on mobile (root cause fixed)

`widget-content` had `overflow: hidden` on mobile. This collapsed the flex chain before browser layout paint, so chart widgets called `clientHeight` at mount time and got `0`. Both TVLiteChart and CandlestickChart rendered invisibly — TVLiteChart's `|| 300` fallback from the previous commit only covered the `createChart()` call, not the container's actual painted height.

**Fix:** Replace `overflow: hidden` + `flex: 1` with `height: 388px` (420px wrapper − 32px header) + `overflow: auto` + `flex: none`. Both chart widgets use `height: 100%` → `flex: 1; min-height: 0` → chart container internally, which now resolves against a real non-zero pixel height.

### 2. Mobile header: all controls visible, version not clipped

The header is a flat `flex` row with no `flex-wrap`. On a phone (375–430px wide), title + status badge + mobile toolbar + version badge all fight for one row. Controls wrap but overflow invisibly; version hangs off-screen.

**Fix:** Enable `flex-wrap` on `.dashboard-header` for mobile. The toolbar (`.layout-controls--mobile`) becomes a full-width second row via `width: 100%`. It's `overflow-x: auto` with hidden scrollbar, so all buttons (layout dropdown, 💾, 🔄/⏸, 🔒, add widget) are reachable by horizontal swipe. Version badge stays in the top row via `margin-left: auto`.

### 3. Resize handle (Bishop's blocking comment, already merged)

Wrapped `.vue-resizable-handle` override in `@media (hover: none) and (pointer: coarse)` so the 44×44px touch target applies only on touchscreens.

### 4. Button touch targets

All `.btn-icon` on mobile get `min-width: 44px; min-height: 44px` (Apple HIG minimum).

---

All 1526 tests pass.

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -1,10 +1,11 @@
 <template>
   <header class="dashboard-header">
+    <!-- Title + connection status (always visible) -->
     <h1 class="dashboard-title">Stock Scanner Dashboard</h1>
     <div v-if="appConfig" class="status success">Connected</div>
     <div v-else class="status error">Error</div>
 
-    <!-- Mobile toolbar: layout selector + lock + widget add -->
+    <!-- Mobile toolbar: layout selector + all controls -->
     <div v-if="appConfig && isMobile" class="layout-controls layout-controls--mobile">
       <!-- Layout load dropdown -->
       <div class="custom-select custom-select--mobile" ref="selectContainerMobile">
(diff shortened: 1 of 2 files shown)


kuhl-haus-bishop (approved):
Chart fix is correct. `overflow: hidden` on the flex child was collapsing the chain before layout paint — `height: 388px; overflow: auto; flex: none` gives charts a real non-zero `clientHeight` at mount time. CandlestickChart handles this natively via ECharts `autoresize: true`; TVLiteChart has its own ResizeObserver. Both work correctly with an explicit ancestor height. Button touch targets, scrollable toolbar, and flex-wrap header all look good. CI green.

One inaccuracy in the PR description worth noting for the record: "Version badge stays in the top row via `margin-left: auto`" is not correct. `app-version` lives after the mobile toolbar in the DOM — so with `flex-wrap: wrap`, `width: 100%` on the toolbar forces it alone onto Row 2, and `app-version` ends up on Row 3. Not a regression (previously it was overflowing off-screen), but if the intent is Row 1 placement, `app-version` needs to move before the toolbar div in the template. Fine to handle in a follow-up.

kuhl-haus-legion:
Reverting — changes made things worse on device. Starting over with a focused fix.

PR #305, opened 2026-06-27
feat(mobile): replace mobile stack with responsive GridLayout for all screen sizes

kuhl-haus-legion:
## Summary

Removes the vertical mobile stack entirely. **GridLayout is now used on all screen sizes.**

### Why

Charts worked in landscape (iPad, phone rotated) because landscape triggers GridLayout — `GridItem` gives each widget a concrete pixel height via its CSS transform, so `height: 100%` resolves correctly inside chart widgets and `clientHeight` at mount is non-zero. The mobile stack never provided that — widgets had `height: auto`, charts rendered at 0px.

The fix isn't patching the stack — it's removing it. The grid already works.

### What changed

**DashboardGrid:**
- Removed `v-if="isMobile"` stack / `v-else` grid split — single `<GridLayout>` for all screen sizes
- Removed mobile-only toolbar; one unified toolbar on all viewports
- `isMobile` is still computed and forwarded to widgets that use it (NewsFeed card mode etc.)
- `gridRowHeight`: 44px on narrow viewports (<640px), 30px on tablet/desktop — taller rows = usable widget heights without manual dragging
- `dashboardColNum` defaults to **2** on narrow viewports (was 12) — widgets span most of the screen width, readable without horizontal scrolling
- Col count input hidden on mobile (not useful at 2 cols; user can still adjust via saved layouts)

**WidgetWrapper:**
- Removed `.widget-wrapper--mobile` fixed height overrides — GridItem handles height

**Tests:**
- Updated mobile layout branch tests: GridLayout always rendered, no mobile-stack
- Updated mobile dropdown tests: unified `.custom-select` selector

### Result

- Charts render on phone in portrait ✓
- Widgets are resizable (grid drag handles work on touch) ✓  
- 2-col default gives a usable layout without any manual setup ✓
- Landscape still works identically ✓
- All 1526 tests pass ✓

### DIFF
--- a/client/src/components/DashboardGrid.vue
+++ b/client/src/components/DashboardGrid.vue
@@ -4,53 +4,8 @@
     <div v-if="appConfig" class="status success">Connected</div>
     <div v-else class="status error">Error</div>
 
-    <!-- Mobile toolbar: layout selector + lock + widget add -->
-    <div v-if="appConfig && isMobile" class="layout-controls layout-controls--mobile">
-      <!-- Layout load dropdown -->
-      <div class="custom-select custom-select--mobile" ref="selectContainerMobile">
-        <div
-            class="select-trigger"
-            @click="toggleDropdown"
-            :class="{ active: isDropdownOpen }"
-        >
-          <span v-if="selectedLayoutName">{{ selectedLayoutName }}</span>
-          <span v-else class="placeholder">Layout</span>
-          <span class="arrow">▼</span>
-        </div>
-        <div v-if="isDropdownOpen" class="select-dropdown">
-          <div
-              v-for="name in savedLayoutNames"
-              :key="name"
-              class="select-option"
-              :class="{ selected: name === selectedLayoutName }"
-          >
-            <span class="option-name" @click="selectLayout(name)">
-              {{ name }}{{ name === defaultLayoutName ? ' ✓' : '' }}
-            </span>
-          </div>
-          <div v-if="savedLayoutNames.length === 0" class="select-option disabled">
-            No saved layouts
-          </div>
-        </div>
-      </div>
-      <button @click="showSaveDialog = true" class="btn-icon" title="Save Layout">💾</button>
-      <button
-          @click="autosaveEnabled = !autosaveEnabled"
-          :class="['btn-icon', autosaveEnabled ? '' : 'btn-icon--inactive']"
-          :title="autosaveEnabled ? 'Autosave ON' : 'Autosave OFF'"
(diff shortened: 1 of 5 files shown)


kuhl-haus-bishop (dismissed):
This is the correct fix. Everything before this was patching symptoms of the mobile stack's `height: auto` problem. GridLayout provides concrete pixel heights via CSS transforms on `GridItem`; the stack never could. Removing it is the right architectural move, not a workaround.

Net diff says it all: −114/+25. Single code path, grid works everywhere, chart rendering is fixed at the root.

A few things I verified:

**`dashboardColNum` default** — `ref(isMobile.value ? 2 : 12)` only sets the initial value for fresh sessions. `loadLayout()` always overrides with `saved.dashboardColNum ?? 12`, so existing saved layouts are unaffected. Fine.

**Alert manager now visible on mobile** — previously gated with `!isMobile`. With the unified toolbar this makes sense; mobile users should be able to manage alerts. Intentional change, correct.

**Test updates** — clean. The `.custom-select--mobile` → `.custom-select` and `.layout-controls--mobile` → `.layout-controls` selector migrations are correct. The `'✓'` → `'(Default)'` change in the default-layout indicator test is accurate — the mobile dropdown used `✓` but the unified desktop dropdown uses `(Default)`. All 1526 pass.

CI green. Good to merge.

oldschool-engineer (approved):

kuhl-haus-bishop (approved):
Clean improvement. The +/- stepper eliminates the NaN edge case entirely, is more touch-friendly than a number input, and the implementation is correct.

Specific things I checked:

- **Boundary handling** — `Math.max(1, dashboardColNum - 1)` and `Math.min(48, dashboardColNum + 1)` are evaluated at click time against the current reactive value, not stale closure captures. ✓
- **Tests** — four new tests in DashboardGridColNum.spec.js covering increment, decrement, min clamp, max clamp, and grid re-render. The old NaN-input test is correctly retired (stepper can't produce NaN). Coverage spec updated to match. All 1526 pass. ✓
- **Accessibility** — `aria-label="Fewer columns"` / `aria-label="More columns"` present. ✓
- **`font-variant-numeric: tabular-nums`** on the display — prevents layout shift as the digit count changes. Nice. ✓

One nit: `.col-step-btn` is 40×40px on mobile, but the other toolbar buttons get 44×44 via the `.btn-icon` mobile media query rule. 4px short of Apple HIG. Not blocking — the col stepper is a secondary control and 40px is close enough in practice — but worth knowing if you're doing an HIG pass.

CI green. Good to merge.

PR #306, opened 2026-06-28
feat(mobile): hamburger button + bottom-sheet menu for portrait phone

kuhl-haus-legion:
## Problem

The toolbar on portrait phone requires horizontal scrolling to reach all controls. Unusable.

## Solution

**Desktop (≥640px):** Unchanged — full inline toolbar.

**Mobile (<640px):** Desktop toolbar is hidden. Replaced with:

- A single **☰ hamburger button** in the header (44px touch target)
- A **bottom-sheet** that slides up from the bottom on tap

### Bottom-sheet contents

| Section | Contents |
|---|---|
| Layout | Full-width layout picker dropdown |
| Actions | 3×3 grid: Add Widget, Lock/Edit, Autosave, Save, Delete, Alerts, Export, Import |
| Columns | Large −/N/+ stepper (56px buttons, edit mode only) |

- Sheet is scrollable, max 90dvh, respects iOS safe-area insets
- Sticky header with ✕ close button
- Tap outside the sheet to dismiss
- Selecting a layout or adding a widget auto-closes the sheet
- Autosave-in-progress shows a pulsing green dot next to the hamburger

## What didn't change

- Desktop toolbar: zero changes
- Grid behavior: zero changes
- All widget logic: zero changes
- 1529/1529 tests pass


oldschool-engineer (dismissed):

kuhl-haus-bishop (changes requested):
Two blocking bugs before this can merge, plus a few non-blocking notes.

---

### 🚫 BLOCKING — two silent no-ops in the action grid

**Vue template inline expressions with `;` are statement bodies, not shorthand calls.** When you write `@click="fn; other = false"`, Vue compiles it as:

```js
($event) => { fn; other = false }
```

`fn` is evaluated as an expression (reads the ref), not called. The second assignment executes. The action silently does nothing.

**Delete Layout button** — `deleteCurrentLayout` is never invoked. Sheet closes, nothing is deleted.

**Export button** — `exportLayouts` is never invoked. Sheet closes, nothing is exported.

Fix — add parens:

```html
@click="deleteCurrentLayout(); mobileMenuOpen = false"
@click="exportLayouts(); mobileMenuOpen = false"
```

The add-widget handler got this right (`(w) => { addWidget(w); mobileMenuOpen = false }`) — apply the same logic to delete and export.

---

### ⚠️ Non-blocking

**`selectContainerMobile` ref likely not wired into click-outside handler**

The mobile sheet dropdown uses `ref="selectContainerMobile"`, but the existing click-outside handler for dropdown dismiss was written against `selectContainer`. Clicking elsewhere inside the sheet won't close the dropdown — the outside-click check won't match the mobile container. Verify `toggleDropdown` / the document click handler covers both refs.

**Import doesn't close the sheet**

Export does `exportLayouts(); mobileMenuOpen = false` — sheet closes. Import just calls `$refs.importFileMobile.click()` — sheet stays open while the file picker appears. Inconsistent. Close the sheet when the dialog opens or in the `@change` handler after `importLayouts` completes.

**No tests for the new surface**

The new conditional rendering and state transitions (hamburger visibility, sheet open/close, action dispatch) have no coverage. The delete/export bugs above would have been caught by tests asserting the expected side effects. Not a merge blocker given 1529/1529, but follow-up issue at minimum.

---

Fix the two no-ops, ready to merge.

kuhl-haus-legion:
Fixed in 4d66159.

**Blocking issues resolved:**
- `deleteCurrentLayout; mobileMenuOpen = false` → `deleteCurrentLayout(); mobileMenuOpen = false`
- `exportLayouts; mobileMenuOpen = false` → `exportLayouts(); mobileMenuOpen = false`

**Non-blocking also addressed:**
- Import button now closes the sheet: `$refs.importFileMobile.click(); mobileMenuOpen = false`

The `selectContainerMobile` click-outside note is filed — will track as a follow-up.

kuhl-haus-bishop (dismissed):
Blocking issues resolved. Approved.

Fix commit addresses all three items: `deleteCurrentLayout()` and `exportLayouts()` parens added, import now closes the sheet. Good.

Remaining open items for follow-up (not blocking merge):

- **`selectContainerMobile` click-outside** — verify the dropdown dismiss handler covers the mobile ref, not just `selectContainer`. If it doesn't, the layout dropdown in the sheet won't close on outside-click.
- **Test coverage** — the new bottom-sheet surface (open/close state, action dispatch) has no tests. The delete/export bugs would have been caught. Follow-up issue.

kuhl-haus-bishop (dismissed):
Approved. `handleClickOutside` correctly covers both refs — all three cases (desktop only, mobile only, both null) behave correctly. Nothing outstanding.

oldschool-engineer (approved):

kuhl-haus-bishop (approved):
LGTM. All issues resolved across three commits. Approved.

